PR #6, opened 2020-01-21
Update uscore

Healthedata1:
update versions and pages


lmckenzi:
I've incorporated the version part of this and also changed the WG to CGP.  However, the page changes are incorrect.  First, for active pages, we don't want to mark them as deprecated.  Second, we need to retain all pages that were previously listed even if they don't exist (those ones should be marked as deprecated).  All resources should be captured using artifacts, not pages and artifact ids should leverage the existing artifact keys where appropriate and only create new ones if there was no equivalent (but mis-named) artifact present.  Finally, the proposed artifact names sort of suck - US Core is redundant because it'll be added to the name.  And suggest "Profile" on the end rather than "StructureDefinition" on the front to be meaningful to most readers.

PR #27, opened 2020-05-05
Adding FHIR-us-dental.xml

dadmbc:
Creating us-dental for Dental Data Exchange FHIR specification (PIID 1553).


dadmbc:
Re-doing through cloning/committing repo

PR #36, opened 2020-06-15
Initial v2 changes

csnewman88:
In V2-core, I deprecated the existing versions, artifact and page as they didn't fit with what the v2MG wanted to start with. If possible, we should delete these altogether to avoid confusion.


lmckenzi:
In the core spec, there's only one artifact and its id is ??.  Can you add the others?

csnewman88:
Updated v2-core to add artifacts and created an Immunization IG spec

PR #37, opened 2020-06-19
Updated artifact and page info

fmckinney:
Updated artifact and page info to match the current IG after applying ballot reconcilation changes


fmckinney:
Hi Lloyd. I attempted to update the Jira info file to match the finished IG, but it failed a check... saying it's an invalid XML file. But XML Spy says it's valid, so I'm not sure what's causing the issue.

fmckinney:
I see I'm going about this wrong.

lmckenzi:
Does it validate when you run locally?

fmckinney:
I was kind of missing the boat… was trying to manually update the file and didn’t get that I needed to upload the file generated by the IG Publisher. I did a separate pull request using the generated file, and after adding the gitUrl attribute it looks like it validates okay now.

PR #44, opened 2020-07-13
Update pages and artifacts for the ImmunizationFHIRDS FHIR IG

csnewman88:
Updates the pages and artifacts for the ImmunizationFHIRDS FHIR IG which is nearing publication.


lmckenzi:
You only need to specify a work group on specific pages or artifacts if it's different from the default.

PR #46, opened 2020-07-20
Add artifacts for SMART Web Messaging

barabo:
The latest build of my IG (in [qa.html](https://build.fhir.org/ig/HL7/smart-web-messaging/qa.html)) suggests that I need to add artifacts in this repo.

I would like to consider this PR draft until I've had a change to get my content right, and also provide documentation in the README.md to help new IG authors get through this process better.


barabo:
I'm not 100% confident in the content of my `xml/FHIR-smart-web-messaging.xml` file, since it's largely just copied from `FHIR-smart.xml` (and I don't understand what the `artifactPageExtension` values mean).

Any suggestions are welcome!

lmckenzi:
The artifactPageExtension stuff isn't actually used yet.  It will eventually be used to help autorecognize that a hyperlink of resource-foo-definitions.html should be considered part of the artifact `resource-foo` and thus auto-categorized accordingly.  If you have no artifacts (as is in the case for this IG), no need to have them at all.  (Does your IG really consist of only one page?)

barabo:
There are a couple pages, actually.

I have https://build.fhir.org/ig/HL7/smart-web-messaging/activity-catalog.html and https://build.fhir.org/ig/HL7/smart-web-messaging/alternatives-considered.html.

Do both of these pages need to be referenced, too?  What would that look like?

```xml
  <page name="activity-catalog" key="activity-catalog" url="smart-web-messaging/activity-catalog"/>
  <page name="alternatives-considered" key="alternatives-considered" url="smart-web-messaging/alternatives-considered"/>
```

PR #51, opened 2020-07-29
C-CDA specs for JIRA Comment Tracking

AStatler:
Added several C-CDA specification documents for C-CDA Comment Tracking.


AStatler:
the pull request failed to find a file that was there. I believe it might be a case issue. It ran successfully in my local build.

PR #75, opened 2020-08-31
Support for Davinci-PDEX-Plan-Net IG

saulakravitz:
1) Renamed the xml for from FHIR-us-davinci-plannet.xml to FHIR-us-davinci-plan-net.xml
2) Replaced the content with the XML generated by the IG publisher.   Several lines are commented out because they caused validation errors.   Only remaining validation errors are as follows, and they may be the entire point of making this change.   

genJson:
      [get] Getting: https://hl7.github.io/JIRA-Spec-Artifacts/SPECS.xml
      [get] To: /Users/skravitz/git/JIRA-Spec-Artifacts/tools/temp/SPECS.xml
      [get] Not modified - so not downloaded
     [xslt] Processing /Users/skravitz/git/JIRA-Spec-Artifacts/xml/_families.xml to /Users/skravitz/git/JIRA-Spec-Artifacts/json/SPECS.json
     [xslt] ERROR: Specification with effective key FHIR-us-davinci-plannet has been removed or changed.  Keys should never change - just change the name.  Keys should also not usually be removed.  Instead, set the 'deprecated' flag to true.  Keys can only be removed if no JIRA tracker references that specification.  If this is the case and the key should really be removed, please coordinate with an administrator.
     [xslt] /Users/skravitz/git/JIRA-Spec-Artifacts/tools/buildSpecJSON.xslt:179: Fatal Error! Processing terminated by xsl:message at line 179 in buildSpecJSON.xslt
     [xslt] Failed to process /Users/skravitz/git/JIRA-Spec-Artifacts/xml/_families.xml

BUILD FAILED
/Users/skravitz/git/JIRA-Spec-Artifacts/build.xml:39: Fatal error during transformation using /Users/skravitz/git/JIRA-Spec-Artifacts/tools/buildSpecJSON.xslt: Processing terminated by xsl:message at line 179 in buildSpecJSON.xslt; SystemID: file:/Users/skravitz/git/JIRA-Spec-Artifacts/tools/buildSpecJSON.xslt; Line#: 179; Column#: -1


saulakravitz:
After 2nd commit, the only error is intentional:

     [xslt] ERROR: Specification with effective key FHIR-us-davinci-plannet has been removed or changed.  Keys should never change - just change the name.  Keys should also not usually be removed.  Instead, set the 'deprecated' flag to true.  Keys can only be removed if no JIRA tracker references that specification.  If this is the case and the key should really be removed, please coordinate with an administrator.


That is precisely what I was trying to do.

lmckenzi:
You can't change the file name.  The file name is the internal key.  Once established, it can't be changed - without a huge amount of manual work in Jira.  The file names and keys are never exposed to the public, so there's no reason to change them.  You can change the display name of your IG at will.

lmckenzi:
You can't remove versions, only deprecate them.

saulakravitz:
Hi Lloyd,
It builds without error, but I thought the whole point was to change the filename."There is no jira specification file defined with the expected name of FHIR-us-davinci-pdex-plan-net in the xml folder of the https://github.com/HL7/Jira-Spec-Artifacts project. This must exist before ballot or publication."

Regards,
Saul
________________________________

saulakravitz:
It builds without error, but I thought the whole point was to change the filename.
"There is no jira specification file defined with the expected name of FHIR-us-davinci-pdex-plan-net in the xml folder of the https://github.com/HL7/Jira-Spec-Artifacts project. This must exist before ballot or publication."

PR #79, opened 2020-10-07
Update FHIR-pddi.xml

rkboyce:
This is to update the JIRA specification for the PDDI implementation guide project.


lmckenzi:
Why are no artifacts listed?

brynrhodes:
Would the profiles be the artifacts?

Also, the default workgroup should be cds I think, not pharma?

lmckenzi:
Yes, and maybe value sets, etc.  The publisher generates  suggested file

brynrhodes:
@lmckenzi , I've looked through the output of the publisher and can't find the suggested file, can you point me to where this would be?

lmckenzi:
It should show up in your root folder - FHIR-uv-pddi.xml

brynrhodes:
I've run the build with the latest publisher and can't find a file anywhere with that name (or anything close to it). Is there some switch I have to set to get the publisher to generate that?

lmckenzi:
You have to be using the HL7 template, which it looks like PDDI is not...

maxsibilla:
FHIR-pddi.xml has been updated with artifacts

PR #80, opened 2020-10-11
Update Pages and Jira data for FHIR-saner

keithboone:


keithboone:
Failing checks are NOT part of this checkin:

[schemavalidate]  /home/runner/work/JIRA-Spec-Artifacts/JIRA-Spec-Artifacts/xml/OTHER-encprs.xml:17:95: cvc-pattern-valid: Value 'Components of ENCPRS FUNCTIONAL PROFILE (Reference) ' is not facet-valid with respect to pattern '[^\s]+([\s][^\s]+)*' for type 'Name'.
[schemavalidate] /home/runner/work/JIRA-Spec-Artifacts/JIRA-Spec-Artifacts/xml/OTHER-encprs.xml:17:95: cvc-attribute.3: The value 'Components of ENCPRS FUNCTIONAL PROFILE (Reference) ' of attribute 'name' on element 'page' is not valid with respect to its type, 'Name'.

lmckenzi:
Sorry about that.  Update it to the new base

keithboone:
Finally ready for merge.

keithboone:
I am unclear as to why this was reporting errors.  The error reports were confusing because they prepended FHIR-saner to the actual keys used in FHIR-saner.xml (those are the key values from the created SPECS.xml file, and I don't understand how those keys are generated.  

In any case, this seems ready to merge.

PR #91, opened 2020-10-28
renamed FHIR-us-davinci-df to FHIR-us-davinci-drug-formulary to align…

saulakravitz:
… with canonical name of IG


lmckenzi:
You can't deprecate the old specification.  You'll lose traceability of all of your old trackers.  The id *cannot* change.  (The same goes for artifacts and pages - if you rename something, you leave the id the same and change the name.)  If the id of your IG has changed from the Jira code, simply add a 'jira-code' parameter in your IG that identifies what the id is in Jira (as per the instructions here: https://confluence.hl7.org/display/HL7/Configuring+Specification+Feedback)

PR #114, opened 2020-11-25
JIRA migration for CDA specs as of 20201125

LisaRNelson:


LisaRNelson (commented):
Josh,  These are the changes we made to fix things discovered by running the .buildlocal.bat file.

hl7webmaster:
invalid due to work made against master branch.

PR #115, opened 2020-11-26
Renamed page, 'missing-data', to 'conformance'

fmckinney:
Sorry, missed one change in my previous request.


lmckenzi:
Did the missing data page get renamed to conformance?  If so, you should keep the original id and just change the page name.  You don't deprecate things just because they've been renamed.

fmckinney:
Ok Lloyd, I'll close this and put the ID back to what it was. Thanks

PR #117, opened 2020-11-29
Updating jira specification file for FHIR-davinci-pdex

ekivemark:
Updating fhir-us-davinci-pdex.xml for v1.0.0 of IG


lmckenzi:
Haven't Patient and Provenance been renamed?  Not sure we should be deprecating the old and adding the new.  Might be better to keep the old key and just use it with the new.

lmckenzi:
What I think you should do is remove the deprecated entry for Provenance and change the new one to say this:
`<artifact id="StructureDefinition/pdex-provenance" key="Provenance" name="PDex Provenance"/>`

I.e. Keep the same key that was there before.

PR #119, opened 2020-12-01
update jira file for us core

Healthedata1:


lmckenzi:
You're deprecating and adding things when you should be renaming things.  In general, you want to keep keys unchanged when renaming artifacts or pages.

PR #129, opened 2020-12-08
Updates

GinoCanessa:
Updated artifacts and pages.
Added version 0.0.1 to match package-list.json


lmckenzi:
You probably want your initial ballot version to be 0.1.0, not 0.0.1

GinoCanessa:
I've updated the ballot version here and in https://github.com/HL7/fhir-subscription-backport-ig/blob/master/package-list.json (hopefully correctly there).  Thanks!

PR #132, opened 2020-12-10
Fix LIVD Jira spec

rhausam:
Remove erroneous FHIR-livd.xml file and commit generated updates to FHIR-us-livd.xml.


lmckenzi:
You should un-deprecate the existing ObservationDefinition key and use that for your ObservationDefinition instance - unless you think comments submitted against the former have nothing to do with the latter.

rhausam:
@lmckenzi 
Looking at the repo history for that file, it was your initial commit of that file on 2/20/2019 that included the deprecated ObservationDefinition key, and that has just carried over since then.  I don't know why it was added by you in the first place (maybe you can recall?).  The generated file from the IG build continues to include it, so I didn't try to do anything about that.  The deprecated ObservationDefinition key seems appropriate for the base spec, but not for comments on the observationdefinition-uv-livd profile in the IG, and it's consistent for the others that will be added for the IG in this PR.  So at this point I think I'm in favor of keeping the existing one deprecated as it is (or removing it if that is possible)?

lmckenzi:
Yes.  All artifacts were marked as 'deprecated' because we wanted the generic FHIR core ones to be replaced with IG-specific ones.  However, where a key already used for a 'core' name corresponds 1..1 with something in the IG, the key should be kept the same.  So best practice is to un-deprecate the key and use it for the IG resource rather than assigning a new key.

rhausam:
@lmckenzi
OK, that's done.

PR #139, opened 2020-12-12
Adjust version

fmckinney:
Adjusted version number and default version for ballot at Lynn's request


fmckinney:
Hi Lloyd. I changed the version number of the Specialty Rx guide from 0.1.00 to 0.1.0 based on a request from Lynn Laakso. I noticed that that caused the jira-spec-artifcact file to go out of sync--which is now showing as a warning during publishing.

The pull request check is flagging the version change as an error, but it looks like an administrator could override it. Just let me know if it's not an okay thing to change, and I can go back to Lynn and let her know. Thanks!

fmckinney:
I'm going to try deprecating the previous version to see if that's the correct way to handle this

PR #141, opened 2020-12-12
Remove 'deprecated' version 0.1.00 - See note

fmckinney:
Hi Lloyd. The update you just merged for me was changing the version from 0.1.00 to 0.1.0 - which Lynn requested in preparation for balloting. I initially created a pull request that just changed the version value without deprecating it, and received an error from the pull check, In response, I manually added an entry deprecating version 0.1.00. (That's what you merged).

But I'm still getting a publisher error... because the entry deprecating the earlier 0.1.00 version isn't in the spec file generated by the publisher... which I probably should have predicted.  So this request removes the entry deprecating 0.1.00. 

Just let me know if there's a different thing I should do to get things back in sync. Sorry for the trouble. Thanks


fmckinney:
Here's the build error, caused when there isn't an entry deprecating the 0.1.00 version. 

> ERROR: Version with code 0.1.00 in specification FHIR-us-specialty-rx has been removed or changed.  **Versions should never change.**  Keys should also not be removed.  **Versions can only be removed if no JIRA tracker references that version.  If this is the case and the version should really be removed, please coordinate with an administrator.**

fmckinney:
Resolved by Lloyd

PR #146, opened 2020-12-14
Adding option for commenting on v2+

csnewman88:
Creating a v2+ spec with an initial version and a basic set of pages


lmckenzi:
I'm reluctant to treat v2+ as a separate specification.  From an ANSI perspective, they're the same thing.  And from most user's perspectives, they're going to be the same thing too...  I think you're better off with a single spec in Jira.

PR #166, opened 2021-01-14
Create OTHER-cql.xml

DonLloyd:
Adding spec file for Clinical Quality Language as part of STU Migration project.


lmckenzi:
Be aware that you still need to update the SPECS-OTHER.xml file to reference this.

PR #167, opened 2021-01-22
Correct bulk data ciUrl and add v1.0.1 and v1.5.0

gotdan:


lmckenzi:
I'll merge this.  Should the http://build.fhir.org/ig/HL7/bulk-data-export/ link go away?  If so, chat w/ Josh and see if he can wipe it.

PR #169, opened 2021-02-04
Changes to V2, Other and CDA files for STU Comment Migration

DonLloyd:
Please see individual comments detailing changes for each of the three commits listed.


lmckenzi (commented):
What is the .project file for?

DonLloyd:
That should not have been included. It was a local file and one I didn't
think I included in the request.

lmckenzi (changes requested):
I'm seeing a bunch of pages called TBD - those shouldn't be there.
Also, why version code of 2.6_R1_STU2.1??  That's super confusing.

DonLloyd:
Hi Lloyd,

The pages with TBD were specifically to address the issue of bringing in
all the existing STU records where we don't know the correct page to
reference. We discussed this on a call a couple weeks ago.

We had also discussed using versions like 2.6_R1_STU2.1 because this is
what the V2 community is used to seeing on their implementation guides. The
Birth and Fetal Death Reporting IG that this sample is from is the STU
Release 2.1 of R1 of the IG specific to Version 2.6. We have other IGs that
are specific only to Version 2.5.1 and we can expect IGs specific to other
versions of V2. Unfortunately, items in the V2 world don't always advance
in a logical straightforward fashion.

Don

lmckenzi:
Ok.  Still failing the checks

DonLloyd:
Hi All,

OK, CDA had incorrectly added cMHAFF to their specifications list. This
will need to be removed by an administrator and the associated
CDA-cmhaff.xml file deleted.

My builds fail locally b/c of this issue. I have updated the SPECS-CDA file
and dropped CDA-cmhaff locally.

Also, I had to update the link to the product brief page in the
OTHER-dam-vr file as I had it incorrect and it matched the url in the
OTHER-cmhaff file.

Don

PR #182, opened 2021-03-10
Updates for STU1.1.0

cspears-mitre:
The team wants to switch to an STU Update 1.1.0 release instead of a 1.0.1 TC release. Also a number of changes were made on code system names and descriptions for HTA.


cspears-mitre:
Will fix and request again.

PR #188, opened 2021-03-29
Updating FHIR-security-ds4p specs

mojitoj:
renaming the file to match the specs id (this used to fail because the build wouldn't allow a number in the key)


lmckenzi:
You shouldn't have 'version' as part of your artifact ids.  The artifact id should remain the same as you update the artifacts.  Version is captured in tracker items separately.  Also, the 'key' should remain the same even when artifact ids do change if the artifact is a logical successor to a previous artifact - we want comments captured against the original artifact to be tracked against the same 'artifact' as the feedback tracked against the new one.  That's important for seeing how content has evolved and for supporting maturity claims.

mojitoj:
@lmckenzi when I left the `Id`s the same I got schema validation errors that the Ids are not unique.

lmckenzi:
The ids in your actual IG should never have versions in them.  If they do, you need to remove them.  If you're getting errors when your IG artifact ids don't have versions, then post a question on Zulip, because that'd be really weird.

The ids in your Jira file should reflect the ids of the resources in your repository now - unless they're an artifact from an old version that's been completely disposed of - i.e. no longer has any equivalent in the new IG.  In that case, it'd be the original artifact id (and the artifact would be flagged as deprecated).  The keys should be the same as the original keys.  Keys never change once established, no matter how many times the artifact id and/or artifact name changes.  We only create new keys when introducing something that's completely net new.  The names in the Jira file should be the current artifact/page name, but you can include older names in brackets to help people find things by the original name if that's helpful.

mojitoj:
@lmckenzi I understood your point but your schema validation does not let me keep the keys and ids the same as the deprecated artifact. Please see the CI log in the last commit. Here is the excerpt:
```xml
    <artifact deprecated="true" id="Observation/example-extension-sec-label-basis" key="example-extension-sec-label-basis" name="example-extension-sec-label-basis" />
    <artifact id="Observation/example-extension-sec-label-basis" key="example-extension-sec-label-basis" name="Simple use of the sec-label-basis extension on an observation"/>
```
here is the error:
```
[schemavalidate] /home/runner/work/JIRA-Spec-Artifacts/JIRA-Spec-Artifacts/xml/FHIR-security-label-ds-for-p.xml:14:176: cvc-identity-constraint.4.1: Duplicate unique value [example-extension-sec-label-basis] declared for identity constraint "artifact-key-unique" of element "specification".
[schemavalidate] /home/runner/work/JIRA-Spec-Artifacts/JIRA-Spec-Artifacts/xml/FHIR-security-label-ds-for-p.xml:14:176: cvc-identity-constraint.4.1: Duplicate unique value [Observation/example-extension-sec-label-basis] declared for identity constraint "artifact-id-unique" of element "specification".
```

lmckenzi:
If you're keeping keys the same, then you don't deprecate the artifact.  You change the id to the new id, change the name to the new name (possibly with key words from old name in brackets) and keep the old key.  The key only appears once and the id only appears once.

mojitoj:
I do want to deprecate the old keys though.

lmckenzi:
Why does your IG have all/most of your artifact ids renamed to have "-new" on the end of them?  The id of an artifact shouldn't change when you change it - even if you make breaking changes.  You only change the id of the artifact if the scope has changed and the name needs to change to reflect the new scope.  "new" isn't useful to an implementer and isn't going to be meaningful in code.  Suggest changing all of your artifact ids back to what they once were.

What I currently see is this:
`    <artifact deprecated="true" id="Observation/example-extension-must-display" name="example-extension-must-display" key="example-extension-must-display" />`
`    <artifact id="Observation/example-extension-must-display-new" key="Observation-example-extension-must-display" name="Simple use of the must-display extension"/>`
    
What I would expect to see is this:
`    <artifact id="Observation/example-extension-must-display" key="example-extension-must-display" name="Simple use of the must-display extension"/>`

That would mean updating your IG to *not* have "-new" on the end of any artifact ids stripping out most of your deprecation assertions and just updating the old artifacts to have the current names.

mojitoj:
I keep saying this @lmckenzi but it's not getting through. I DO want to change the keys because they were chosen poorly and the only way to change them is to deprecate the artifact according to the docs. I completely understand your point about keeping traceability with old artifacts but in this case, having organized and proper strings for IDs and keys is more important to me in order to make sure this document is maintainable in the long run. 
Again, I understand your point but I do think our need to clean up the IDs and keys (which were chosen poorly and inconsistently in the initial version of this document) is more important. We should not have to be stuck with poor identifiers and keys forever. 
Since you and I seem to disagree about this I do want to escalate this to have other people comment on this.

lmckenzi:
You *cannot* change keys once established.  The keys aren't relevant - no one sees them.  They are not visible in Jira.  The key for CapabilityStatement is still "conformance" because that's what it once was.  The same is true of other resources.   What matters about keys is that they stay the same, not what they say.  You're free to change ids if need be (though don't add something like '-new' to them).  You're certainly free to change names.  But keys can't be changed, no matter how inappropriate they might have been.  They matter only inside the Jira tooling, so they shouldn't matter to you or anyone else,

lmckenzi:
Looks good and I will merge.  Only remaining issue I can see is that you're declaring a work group on one of your pages, and you only need to do that if the work group for the page differs from that of the overall spec.

mojitoj:
Thanks @lmckenzi. I will fix that in our next update.

PR #211, opened 2021-04-12
Updates to us-military-service

swmuir:
Updates to us-military-service


lmckenzi (commented):
You're deprecating a *lot* of artifacts.  None of those artifacts have any correlation with new artifacts?  Not clear what's driving the radical change here.

PR #213, opened 2021-04-12
Update current/ci urls for ade and med-admin IGs

minigrrl:


lmckenzi (changes requested):
The CI build shouldn't point to a branch.  Also, a URL should never include a filename - it should be the base string that we check for in full URLs.

PR #220, opened 2021-04-23
Update for STU Update Peer Review

cspears-mitre:


lmckenzi (commented):
I thought we *weren't* introducing 1.0.9?

Also, is the move from ClaimSubType to InstitutionalClaimSubType a true deprecate and add, or can we better handle that as a re-name and re-scoping?  (Base question is "should there be traceability in terms of feedback received on ClaimSubType to the new InstitutionalClaimSubType?)

cspears-mitre:
We are not introducing a 1.0.9. This is updates without 1.0.9. The change in name was to correct a typo on my part ClaimSubType (the WG voted on the Institutional name) and never existed in a published version. We can handle as a rename (no scope change meant). I do not think there is any need for traceability here as no one has used it yet. What do I need to do to get this done as quickly as possible?

lmckenzi:
This pull request *is* defining version 1.0.9 - so you should remove that.
If it was a typo, keep the old key and just change the name and id, don't deprecate and add.

PR #225, opened 2021-06-02
Update for Publish 1.1.0 request

cspears-mitre:


lmckenzi:
Why is there a '9' in the key for change notes?

cspears-mitre:
> Why is there a '9' in the key for change notes?

It took the file name. The 9 in the file names was to enforce a page order.

lmckenzi:
Sigh.  That was never a great idea.  Ok, I'll approve as is.

cspears-mitre:
> Sigh. That was never a great idea. Ok, I'll approve as is.
Thanks.
I was just keeping it with the convention of this IG thus far. I am not introducing it to other IGs I am working on.

PR #229, opened 2021-06-16
new errata release version

brettmarquard:
Need 2.1.0.7 to support marking items applied


lmckenzi (commented):
Don't you want to mark 2.1.0.6 as deprecated?  (i.e. you won't take change requests against that release anymore)

brettmarquard:
Hmm - I think once we 'publish' 2.1.0.7 we would deprecate .

lmckenzi:
Ok, so you'll do that when you change the default version to 2.1.0.6 - and you'll remember to do so? :)

PR #232, opened 2021-06-24
update to v 4.0.0

Healthedata1:


lmckenzi (commented):
Are you still accepting changes against all previous versions, or should some of them be marked as deprecated (meaning that issues can't be raised against them anymore)?

Healthedata1:
Will need to confirm with CGP/US Realm, but our inclination is to deprecate all but the most recent 3.1.1.  Should the ballot versions need to be deprecated as well as a matter of course?

lmckenzi:
Yes.  Anything you don't want to accept subsequent change requests against should be flagged as 'deprecated'.

Healthedata1:
Done

Eric M Haas, DVM, MS
Health eData Inc
35 Crescent Avenue, Sausalito, CA 94965
707.227.2608|Skype: haas.eric1
***@***.***

PR #233, opened 2021-06-28
Updates for VRDR

AbdulMalikShakir:


lmckenzi (commented):
Should some of the versions - particularly 0.1 be marked as deprecated?

AbdulMalikShakir:
> Should some of the versions - particularly 0.1 be marked as deprecated?

All deprecated artifacts and pages have been marked deprecated="true". Should versions be deprecated given the need for version history?

lmckenzi:
Versions that are no longer supported (i.e. that you don't want to accept change requests against and that you won't be making fixes to) should be deprecated.  They'll still remain for historical trackers, but won't be available for new ones.

AbdulMalikShakir:
> Versions that are no longer supported (i.e. that you don't want to accept change requests against and that you won't be making fixes to) should be deprecated. They'll still remain for historical trackers, but won't be available for new ones.

I've deprecated all but the STU1 version and the current CI build version.

lmckenzi (changes requested):
Why do 1.1.0 and 1.0.0 have the same website?  Typically STU updates result in a new website?

AbdulMalikShakir:
> Why do 1.1.0 and 1.0.0 have the same website? Typically STU updates result in a new website?

The STU1.1 has not made it through publication yet. I suspect that when it does the wedsite address will change. Should I change it here now in anticipation?

lmckenzi:
Yes please.

AbdulMalikShakir:
> Yes please.

Done. The URL for version 1.1.0 now points to STU1.1

lmckenzi (approved):

PR #235, opened 2021-06-30
roll-back defaultVersion

lorifourquet:
use the exported defaultVersion


lmckenzi (commented):
I don't understand.  Why would 1.1.0 not be the default version?  It is (or very soon will be) the official release, no?  Isn't that what you'd expect most people to be providing feedback against?
Also, do you intend to ever make technical corrections against 1.0.0 now that 1.1.0 is going out?  If not, you should mark it as deprecated.  You should definitely mark 0.1.0 as deprected.

I expect the 'default' is wrong because the releases are in the wrong order in your packagelist.json file.

Finally, there's a whole bunch of whitespace changes.  Those are probably going to cause validation issues against the generated version.  Did you format the output with some sort of tool?

lmckenzi (commented):
Are you going to mark 0.1.0 (and maybe 1.0.0) with deprecated=true?

lmckenzi (changes requested):
This file looks like it's been completely reformatted.  That's going to cause comparison problems.  Are you able to edit the original file with something that doesn't change formatting?  Also, the order of the versions has been changed - and doesn't seem to be "in order" anymore...

lmckenzi:
Order still wasn't quite right.  I adjusted and committed manually.  Can you see if it works now?

PR #237, opened 2021-07-09
Configuration for Radiation Dose Summary IG

aboufahj:
The related IG CI Build is under: https://build.fhir.org/ig/HL7/fhir-radiation-dose-summary-ig/
The related IG project summary is under: http://www.hl7.org/special/Committees/projman/searchableProjectIndex.cfm?action=edit&ProjectNumber=1692


lmckenzi (changes requested):
Initial version shouldn't be 0.1, should be 0.1.0.  Also, location should be the version-specific permalink.  Finally, default version should be the 'current' version.

lmckenzi:
It doesn't look like these commits addressed the issues I raised?  You'll need to adjust your packagelist.json to have the appropriate version declarations.

lmckenzi:
You can't remove the 0.1 version.  (You can deprecated it though)

PR #242, opened 2021-07-21
Adding US Pathology data sharing and UV Pathology Common Library specs

dadmbc:


lmckenzi (changes requested):
There's no URL specified for the 0.1.0 version and there's no 'current' version defined.  (Both should come from your package-list.json.)

dadmbc:
Both IGs appear to have a path defined for version 0.1.0:
UV: "path": "http://hl7.org/fhir/uv/cancer-pathology-common-libraries/2021SEP"
US: "path": "http://hl7.org/fhir/us/cancer-reporting/2021SEP"

Could you please confirm that 'current' should always be the CI build site (ie never the Ballot version)? If so, I need to update both package-list.json to reflect this. Currently, both have the STU1 version as current.

lmckenzi:
'current' should always be the CI build.  The path for the version isn't propagating for some reason - all I'm seeing is <version code="0.1.0"/>

dadmbc:
I'm thinking it might not exist yet; this is (or will be) the HL7 ballot site. When/how does this get created?

lmckenzi:
The path for the ballot release won't exist yet - but it should be listed in the package-list.json when you're prepping for ballot and should be in the JiraSpecArtifacts file so that once publication occurs, everything works properly.  The URL for the ballot will be the canonical URL with "2021Sep" stuck on the end.

dadmbc:
Understood. I've updated the current version to be correct on both package-list.jsons. What else do I need to do in order to successfully create the jira specs for these new IGs before tomorrow's content freeze (given the ballot url has not been created yet)?

lmckenzi:
If you regen the IGs, they should produce new Jira-Spec-Artifact candidates - update your pull requests to reflect those.

dadmbc:
ok

PR #243, opened 2021-07-22
Add SHR Vaccination IG

masnick:


lmckenzi (changes requested):
One small change.  The path for 0.6.0 should be http://hl7.org/fhir/uv/fhir-shc-vaccination/2021Sep, as that's presumably where the frozen snapshot of that release will be hosted.

masnick:
> One small change. The path for 0.6.0 should be http://hl7.org/fhir/uv/fhir-shc-vaccination/2021Sep, as that's presumably where the frozen snapshot of that release will be hosted.

Thanks, @lmckenzi -- fixed this.

lmckenzi (approved):

PR #245, opened 2021-07-26
Add in ICSR AE Reporting FHIR IG

jduteau:


lmckenzi (changes requested):
Initial version should be 0.1.0.  version should specify the url of the expected URL for the ballot.

lmckenzi (changes requested):
URL for 0.1.0 should be the location the ballot will be located (presumably a 2021Sep location).  It shouldn't point to the CI build.  You should have a 'current' version that points to the CI build.

lmckenzi (approved):

PR #247, opened 2021-07-29
Update FHIR-saner spec

DavidPyke:
Match to pre-pub version

### DIFF
--- a/xml/FHIR-saner.xml
+++ b/xml/FHIR-saner.xml
@@ -20,7 +20,8 @@
      <artifact id="CodeSystem/IHE-ADX-agerange" key="CodeSystem-IHE-ADX-agerange" name="Age Range Coding System"/>
     <!-- Renamed MeasurePopulationAlternateCriteria to MeasureAlternateCriteria -->
      <artifact deprecated="true" id="MeasurePopulationAlternateCriteria" key="MeasurePopulationAlternateCriteria" name="MeasurePopulationAlternateCriteria"/>
-     <artifact id="StructureDefinition/MeasureAlternateCriteria" key="StructureDefinition-MeasureAlternateCriteria" name="Alternate criteria for performing a measure"/>
+	 <artifact deprecated="true" id="StructureDefinition/MeasurePopulationAlternateCriteria" key="StructureDefinition-MeasurePopulationAlternateCriteria" name="Alternate criteria for performing a measure"/>
+     <artifact id="StructureDefinition/MeasureAlternateCriteria" key="StructureDefinition-MeasureAlternateCriteria" name="Alternate criteria for performing a measure."/>
 
     <!-- deprecated StructureDefinition/AuditEvent{Create,Delete,Read,Search,Update} and StructureDefinition/audit-event-read and StructureDefinition/audit-event-write -->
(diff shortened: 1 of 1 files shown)


lmckenzi:
You're deprecating a *lot* of stuff and adding a lot of stuff - the stuff added has no correlation with the stuff being deprecated?

DavidPyke:
I worked from the files created by the publisher.  As there are hundreds of pages in this spec, doing a line by line is time consuming and difficult.  Since we're going for STU 1 publication, I'm hoping that this will put a line in the sand for new specs.

DavidPyke:
The ballot process changed a lot of the structure and order of the pages so much of the changes are difficult to minimize in the spec file

lmckenzi:
If an artifact in the new version of the IG is the same as one in the previous version, the key should remain unchanged (you're free to change the id and title).  Yes, it's a lot of work.  But if you don't, then there's no continuity between feedback received on the original release and the new release.  That traceability matters (both so users can see how things have evolved and why, but also for tracking FMM - how many substantive changes have happened to particular artifacts based on feedback).

DavidPyke:
This is the original release, the ballot release and CI "releases" are unlikely to be a concern into the future.  I'm quite happy to leave them behind as we have had extensive feedback throughout the cycles and are setting the flag as to SANER V1.  Having people follow the breadcrumbs isn't that important, at least to us.  

As for the keys changing, they were all set by the publisher, if that tool can maintain the keys but change id, etc. this wouldn't be as difficult

lmckenzi:
If the id remains the same or the name remains the same, the tool will keep the key the same.  If both have changed, the tool depends on a human realizing "that's not a deprecation and an addition, that's a change" and adjusting the file accordingly.  By not adjusting the keys, you're saying "this past ballot won't count from a maturity perspective" - that seems an odd choice to make for what shouldn't be more than an hour's work...

DavidPyke:
I don't understand what "this past ballot won't count from a maturity perspective" means

keithboone:
I've done my best to comment on the various changes made and clean up white-space to match the original so that you can better see the differences and understand the changes.

Sadly, the original has use of tabs that Eclipse keeps changing on me even after telling it not to.

You can just tell git to ignore whitespaces in the diff.
image

lmckenzi:
The maturity criteria for level 3 include "at least 10 distinct implementer comments recorded in the tracker drawn from at least 3 organizations resulting in at least one substantive change".  Those comments are tracked by looking for tracker items associated with the artifact.  If you lose the connection to past tracker items, they won't count.

For the stuff you're just yanking, no issue.  However, for the stuff that you're renaming, change your file to retain the original key.

DavidPyke:
So, if the key is deprecated, the jira ticket against the IG will disappear?  That doesn't make sense to me.

lmckenzi:
If a key is deprecated, that says "this conceptual thing no longer exists and the tickets against it are no longer relevant.  If a concept has simply been renamed or the id has changed, then the key should remain the same and the Jira tickets will propagate to now be associated with the new name.  Simple example - all of the Jira tickets against 'Conformance' in STU1 are now tied to CapabilityStatement.

DavidPyke:
Okay, then this should all be right for maintaining history

PR #260, opened 2021-08-09
Update FHIR-us-sirb.xml

jdtopping:
Enclosed new page item in round brackets

### DIFF
--- a/xml/FHIR-us-sirb.xml
+++ b/xml/FHIR-us-sirb.xml
@@ -13,9 +13,17 @@
 <artifact id="Questionnaire/POPULATE-sirb-determination-letter-questionnaire" key="Questionniare-POPULATE-sirb-determination-letter-questionnaire" name="POPULATE Determination Letter"/>
 <artifact id="ValueSet/temporarycodes" key="ValueSet-temporarycodes" name="Temporary Codes"/>
 <artifact id="ResearchStudy/ResearchStudy2858" key="ResearchStudy-ResearchStudy2858" name="Research Study"/>
+<artifact id="Questionnaireresponse/INITIATE-QR" key="Questionnaireresponse-INITIATE-QR" name="Initiate Study Questionnaire Response"/>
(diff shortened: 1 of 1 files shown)


lmckenzi:
Putting brackets around the qr 'page' won't help.  There are already artifacts for each qr.  You shouldn't have a 'theoretical' page for qrs when you have artifacts for them too.  (In general, you only want one way to provide feedback for a given page in the IG.)

jdtopping:
I didn’t make artifacts for the responses, just the questionnaires.  I was trying to group the feedback for the responses together since there’s likely to be less on the responses themselves.  So they should all be their own artifacts independently?

lmckenzi:
If they're not artifacts listed in the IG, they won't show up at all.  And
I absolutely see them as artifacts in the IG...

*Lloyd McKenzie*, P.Eng. (he/him)
Senior Consultant, Information Technology Services
Gevity Consulting Inc.

 E: ***@***.***
M: +1 587-334-1110 <1-587-334-1110>
W: gevityinc.com

*GEVITY *

*Enabling a healthier world *

CONFIDENTIALITY – This communication is confidential and for the exclusive
use of its intended recipients. If you have received this communication
through error, please notify the sender and delete the message without
copying or disclosing*.*

REPRESENTATION –  Unless explicitly stated otherwise, the opinions and
positions expressed in this e-mail do not necessarily reflect those of my
employer, my clients, nor the organizations with whom I hold governance
positions

jdtopping:
Lloyd-

It looks like I am still getting the warning about the Jira artifact or version tags not matching.  I deprecated the random page and added all the Questionnaireresponses as artifacts.  Any thoughts?  Thanks for your help as always.

James

lmckenzi:
Issue is that the keys had been changed when the ids were corrected.  The keys needed to be left as they were.  (The fact they have spelling/capitalization issues doesn't matter.)  Fixed with a separate commit.

jdtopping:
It looks like I was missing a number of other rows as well, so thanks for the help.

James

PR #261, opened 2021-08-09
Update FHIR-pocd.xml

stefan-karl:
Update FHIR-pocd.xml with unmodified file as generated by the IG publisher (after name changes to distinguish CodeSystem and ValueSet).


stefan-karl:
The IG publisher removed one line that I added manually because otherwise .buildlocal.bat failed:
<artifact name="Observation" key="Observation" deprecated="true"/>
Is it really needed?

lmckenzi:
Was there an Observation key in the original version?  If so, the publisher shouldn't remove an update that deprecates it...

stefan-karl:
There is an Observation profile with key="StructureDefinition-Observation" in FHIR-pocd.xml.
The entry with key="Observation" was deprecated from beginning.

lmckenzi:
This is extremely stale, so closing

PR #264, opened 2021-08-09
updating dental-data-exchange

dadmbc:
Re-organizing and updating deprecated pages and artifacts


dadmbc:
"ERROR: Version with code 0.0.1 in specification FHIR-us-dentalreferral has been removed or changed.  Versions should never change.  Keys should also not be removed.  Versions can only be removed if no JIRA tracker references that version.  If this is the case and the version should really be removed, please coordinate with an administrator." 

Version 0.0.1 has an incorrect URL attribute. Proposing the three versions for this specification should be:
  <version code="current" url="http://build.fhir.org/ig/HL7/dental-data-exchange" />
  <version code="1.0.0" url="http://hl7.org/fhir/us/dental-data-exchange/STU1" />
  <version code="0.1.0" url="http://hl7.org/fhir/us/dental-data-exchange/2020Sep" />

Also, some updated to the <specification> attributes are required for accuracy as well (compare below with existing):
<specification xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
  gitUrl="https://github.com/HL7/dental-data-exchange" url="http://hl7.org/fhir/us/dental-data-exchange"
  ciUrl="http://build.fhir.org/ig/HL7/dental-data-exchange" defaultWorkgroup="pie" defaultVersion="1.0.0"

lmckenzi (commented):
There are quite a few 'deprecated' artifacts - are these artifacts that have been renamed, or are they artifacts that have been removed and no longer have any equivalent in the IG?  If the former, then the old key should be put on the new artifact so that comments made against the artifacts with the original names will carry forward to be associated with the renamed artifacts.

dadmbc:
These are true deprecated as a result of implementing ballot comment dispositions.

PR #269, opened 2021-08-10
Updating ballot Url from /2021-SEP to /2021Sep

dadmbc:
Updating ballot URL from 2021-SEP to /2021Sep


dadmbc:
Automated checks usually don't take this long, is there something I can do to nudge it?

lmckenzi:
I nudged, but that doesn't seem to have helped.  Have reached out for support

lmckenzi:
There was something wonky with Git.  Can you make a whitespace change and push the update to force it to retry?

dadmbc:
Seems to have worked, ty

PR #270, opened 2021-08-10
eyecare fix canonical url

oridashi:


oridashi:
a correction to canonical URL was needed

oridashi:
can I get the key change approved for this; was to match the pattern for corrected canonical URL - was uv-eyecare changed to 'eyecare'; please advise if this is an issue i could change back - no Jira tracker history as this is new IG

lmckenzi:
Once the key is set to eyecare, we can't easily change it.  Because it's different from your code, you'll need to set the jira-code parameter in your IG, as described here: https://confluence.hl7.org/pages/viewpage.action?pageId=76153131#ConfiguringSpecificationFeedback-Filetomodify

oridashi:
thanks @lmckenzi I have returned to orginal key and updated IG: advice is much appreciated

PR #273, opened 2021-08-11
Military service history

IoanaSingureanu:
Updated "ballotUrl" and the canonical URL to include the realm.


lmckenzi (changes requested):
The CI build *should* be https://build.fhir.org/ig/HL7/fhir-military-service, not https://build.fhir.org/us/HL7/fhir-military-service.  The first resolves, the second doesn't.  Do you have the CI-build URL incorrect in the package-list.json file?

PR #276, opened 2021-08-12
Update FHIR-us-military-service.xml

IoanaSingureanu:
Removed the prefix "fhir-" from the ballot and canonical url.
Unfortunately the jira-new specifies "defaultGroup=''" which will continue to create the jira spec mismatch warning.


IoanaSingureanu:
This should solve my all my problems except for the the empty "defaultWorkGroup".

PR #284, opened 2021-08-19
Changes to FHIR-us-covid19.xml file

nrdavis1:
Made modifications according to Lloyd McKenzie's comments and applied fix to pass build.


lmckenzi (changes requested):
Still missing the version url and the 'current' version.

nrdavis1:
@lmckenzi Can you give mean example of what the version URL looks like as well as the current version?  I'm comparing what I have to other files and don't see any difference.

lmckenzi:
Your versions should look something like this:
<version code="current" url="http://build.fhir.org/ig/HL7/cdisc-map"/>
<version code="0.1.0" url="http://hl7.org/fhir/uv/cdisc-mapping/2021Jan"/>

nrdavis1:
@lmckenzi I don't see the examples you provided.  Did you attach them?

lmckenzi:
Sorry.  Looks like Git doesn't like pasted XML.  Updated my post above to escape the open-brackets.

PR #301, opened 2021-09-30
updates to accommodate the renaming of the "must-display" extension to "display"

mojitoj:


lmckenzi (changes requested):
If you're renaming an artifact, don't deprecate the old one and add a new one.  Keep the key the same and just change the id and/or name.

lmckenzi (approved):

PR #303, opened 2021-09-30
Add IPA to spec list

vassilpeytchev:
Adding IPA to the specification list.


lmckenzi (changes requested):
<version code="1.0.0" url="http://www.hl7.org/fhir/uv/ipa/STU1"/>
should omit the "www."

vassilpeytchev:
Fixed. Should I also change it for IPS (which is where I copied it from)?

lmckenzi:
> Fixed. Should I also change it for IPS (which is where I copied it from)?

Please :)

lmckenzi (approved):

PR #304, opened 2021-09-30
Updating dentalreferral

dadmbc:


dadmbc:
@lmckenzi I'm hoping to update the dental referral Jira spec with the file that was provided to Sarah on yesterday's FMG call. However, this file is not valid- as it has 3 duplicate values on deprecated artifacts. If I delete the deprecated problematic artifacts, the issue is not expected to resolve. Please advise.

lmckenzi (commented):
And none of the deprecates are 'rename' situations?

dadmbc:
2 artifacts are renaming scenarios (the rest are truly deprecated from the IG): Referral Note and Consult Note StructureDefinitions.

PR #317, opened 2021-10-26
carin-dic

jdjkelly:
Per https://github.com/HL7/carin-digital-insurance-card/pull/14 this is the new JIRA config for CARIN Digital Insurance Card cc @cspears-mitre  @cillekisselwatkins


jdjkelly:
@lmckenzi we seem to be still generating a diff on HL7/carin-digital-insurance-card for the JIRA config. The diff is the same as this build failure: https://github.com/HL7/JIRA-Spec-Artifacts/runs/4015537366?check_suite_focus=true

The JIRA template script seems to be generating 3 name collisions on the project for the JIRA artifact keys when, from what we can tell, it shouldn't be. Do you have any insight here?

cc @cspears-mitre you may also know?

lmckenzi:
I'm not understanding what you mean.  Jira-Spec-Artifacts seems to be building fine.  If you're getting a diff in your 'new' and 'old' files send me a copy of both files.

jdjkelly:
Hi @lmckenzi apologies for the ambiguity above, I posted a more detailed report with diffs here: https://chat.fhir.org/#narrow/stream/179252-IG-creation/topic/JIRA.20Spec.20name.20artifacts.20collission

PR #327, opened 2021-10-28
Blew away fork again and copied in new file. Now I get no errors.

brianreinhold:
I don't understand it, but I cloned the unforked master, copied in my replacement file, and the build worked with no errors. When I did the same with the forked repo, I got warnings and an error. So I blew away the forked repo again, forked again, and copied in the replacement file. Now it worked here as well.

I'm at a total loss as none of the other files in the fork were ever touched.


brianreinhold (commented):
The commit message is out to lunch. Don't understand why it is what it is. It is not the commit message I made today. It was an old message.

PR #339, opened 2021-11-08
Merge HL7 JIRA-Spec-Artifacts master and add new jira spec for pacio-rt

cwanmtr:
Add new jira spec for pacio-tr


lmckenzi (changes requested):
You need to modify SPECS-FHIR.xml so that this spec file actually gets processed.

cwanmtr:
> You need to modify SPECS-FHIR.xml so that this spec file actually gets processed.

Done. Thanks for pointing this out.

lmckenzi (approved):

PR #345, opened 2021-11-12
Da Vinci Formulary STU2 Ballot

cspears-mitre:


lmckenzi (commented):
Some of the adds and deprecates look like they should actually be renames?  If all you've done is change the id and name but the 'concept' is the same, then please keep the key the same and just change the id and name.

cspears-mitre:
Spec was largely rewritten. Will look closer and make a separate request.

PR #356, opened 2021-11-23
Update Risk Adjustment to match changes made to the IG for ballot

rob-reynolds:


rob-reynolds:
Fixing bug in several artifact names.

PR #357, opened 2021-11-23
update pct spec

yuliumitre:
Update PCT spec file.


yuliumitre:
The build failed due to the following error:

ERROR: Version with code 1.0.0 in specification FHIR-us-davinci-pct has been removed or changed.  Versions should never change.  Keys should also not be removed.  Versions can only be removed if no JIRA tracker references that version.  If this is the case and the version should really be removed, please coordinate with an administrator.

It is caused by the mismatch of the version between the initial placeholder FHIR-davinci-pct.xml with the generated one. 1.0.0 was used in the template but the real version should be 0.1.0. Is there a way to force this change? E.g. force merge. Thanks.

yuliumitre:
@lmckenzi Hi, Lloyd, this is a new IG and there is no existing spec. Can you help resolve the build error? Thanks.

yuliumitre:
@lmckenzi Hi, Lloyd, would you able to help merge this PR? Thanks.

PR #361, opened 2021-11-27
xml/FHIR-us-davinci-df.xml

cspears-mitre:
This IG has gone through a significant rewrite. All deprecated items have been reviewed. The same key was maintained where possible and where keys could clearly be connected to the underlying spec concepts. 
Some deprecated may look like simple renames, but there are significant differences, such as the concept (profile) was split into two different profiles using completely different resources. Maintaining the same key will cause confusion to voters and to maintainers of the spec over time.


cspears-mitre:
Github error. Intermittent service issues. Will retry.

PR #379, opened 2021-12-03
Updates to Marketplace IG

swmuir:
Added new keys


lmckenzi (changes requested):
Once keys are established, you should keep them and not change them.  Even if you change the name, the key should remain the same.  Don't deprecated the old keys and add new rows.

lmckenzi (approved):

PR #395, opened 2021-12-07
JIRA spec artifacts for Care Team DAM

MLD1995:


lmckenzi (commented):
Is it your intention that feedback on the DAM can't be differentiated into different sub-parts of the DAM?  I.e. no distinct sections, artifacts or anything else?

MLD1995:
Yes. It's a

> Is it your intention that feedback on the DAM can't be differentiated into different sub-parts of the DAM? I.e. no distinct sections, artifacts or anything else?

Yes. It's an extremely short document without any sub-parts. When I was walking through the process of creating the JIRA spec file with the project team the agreed they didn't have any sub-parts.

PR #418, opened 2022-01-28
Update to cimi-vital-signs spec file

nrdavis1:
Updated based on changes from ballot reconciliation.


lmckenzi (changes requested):
It looks like you're deprecating and adding things that have merely been renamed.  That's the automated logic for generating the file, but where what's happening is a rename, you should keep the old key and merely change the identifier and/or name, don't do a delete and add.

PR #421, opened 2022-01-31
Update to FHIR-us-vitals Jira-Spec-Artifacts file

nrdavis1:


lmckenzi (changes requested):
It looks like there a are a number of places where an artifact is being deprecated and a new one added when instead the existing artifact should have been renamed.  I.e. keep the old key and use the new id and title.

lmckenzi (approved):

PR #462, opened 2022-03-17
first stab at updating fhircast' file

isaacvetter:
as part of moving FHIRcast to an IG.


isaacvetter:
@lmckenzi -- Mr McKenzie, this PR is now passing the automated checks! Would you be so kind as to merge it as your earliest convenienece?

lmckenzi:
I *can*.  However, I see that a bunch of your keys reflect section hierarchy.  Be aware that once the key exists, it's immutable.  Even if the section moves, if it's the same content, the key needs to stay the same.  Do you really *want* to put section numbers in the keys?

isaacvetter:
Lloyd,

Thank you, you're exactly right. Your suggestion is implemented, I also cleaned up the page key names in general. Any other feedback along these lines?

Isaac

lmckenzi:
Only other feedback is that the 'dropdown' in Jira will be in the order you list things in.  Most of the time you want it to be alphabetical because that makes it easier for people to scroll and find.  Though sometimes, you might want to put super-common things up at the top.  I'll wait for your confirmation before merging.

isaacvetter:
@lmckenzi - thank you, alphabetized keys, with some meangingful exceptions. As soon as you have time, please do merge.

PR #489, opened 2022-03-31
FHIRcast - last ditch effort

isaacvetter:
specification.url should be build url
STU1 and STU2 urls must be on hl7.org!


isaacvetter:
Deferring to Tony's #490

PR #537, opened 2022-07-26
pacio pfe initial jiraspec file

karlnaden:


lmckenzi (changes requested):
0.1 doesn't seem aligned with current versioning guidelines - think it should be 1.0-ballot

karlnaden:
updated

PR #547, opened 2022-07-29
update version to 1.1.0-ballot

fmckinney:


fmckinney:
Replaced with corrected update

PR #565, opened 2022-08-04
national directory IGs

smahoney10:


lmckenzi (changes requested):
Artifact pages should not be listed under 'pages'.  Only the pages that correspond to non-generated pages and base pages like ToC and artifacts list should be in the list.

smahoney10:
fixed deprecation erros

PR #575, opened 2022-08-10
New version of Pharmacy Templates being balloted.

jduteau:


lmckenzi (changes requested):
The ballot version should be something like 1.2.0-ballot or 2.0.0-ballot (depending on what you expect the final version to be)

lmckenzi (approved):

PR #581, opened 2022-08-18
New MDI Jira Spec update

DianaWright61:


DianaWright61:
"MDI FHIR IG Jira Spec update" failed because of extra white spaces, I think. "Corrected Jira Spec - MDI" maybe has fixed that.

PR #586, opened 2022-08-24
Update to FHIR-us-vitals spec file

nrdavis1:


nrdavis1:
The check failed due to an error in the CDA-pharm file.  Why am I getting the error?  I didn't touch that file.

PR #594, opened 2022-09-03
re-assessment timepoints STU 1 publication prep

karlnaden:


lmckenzi (commented):
You've deprecated a lot of artifacts - are all of them truly removed, or are some of them re-names?  If they're re-names, you should keep the old key and just update the id and/or artifact name.

karlnaden:
I did do a lot of renaming of examples to better organize them.  I tried making adjustments to keep the original ID, but they don't seem to have worked, so I must not be doing something right.  I assumed there might be some matching going on behind the scenes so that if I updated the keys in the FHIR-us-pacio-rt.xml file, it would find those elements using the name or id fields and keep the key, but looks like my changes just got overwritten instead when I went to re-run the publisher. Any pointers on how to rename / re-id correctly without changing the key?

lmckenzi:
The trick with renames is to retain the old key, but specify the new id and name.  If you do that, then you won't need to deprecate the old instance.  (The algorithm that generates the 'candidate' file will always have a row to deprecate the old and add the new because it doesn't know any better - but you can always override that.)

karlnaden:
Does that mean I (or whoever builds the IG) would need to manually adjust or ignore changes to the file before building? Seems like a lot of manual effort to avoid deprecation of resources that don't have any comments against them in JIRA.  Or am I still missing a way to make my changes sticky so that they don't have to be redone?

lmckenzi:
Once you've changed a key to a new id/name that'll be sticky.  You'll only need to undertake manual effort again if/when you rename a page or change the id of a resource again.

karlnaden:
The mechanics are not working as I thought they would, so let's go through a concrete example

balloted spec file has this entry
<artifact id="Condition/BSJ-Diagnosis1" key="Condition-BSJ-Diagnosis1" name="BSJ-Diagnosis1"/>

I updated the id of this example resource to be clearer from BSJ-Diagnosis1 to RT-SNF-Diagnosis-CI. That resulted in the following two lines in the spec file when I built the IG

<artifact deprecated="true" id="Condition/BSJ-Diagnosis1" key="Condition-BSJ-Diagnosis1" name="BSJ-Diagnosis1"/>
<artifact id="Condition/RT-SNF-Diagnosis-CI" key="Condition-RT-SNF-Diagnosis-CI" name="RT-SNF-Diagnosis-CI"/>

However, since this is a rename, I really want to avoid the deprecation, which I think you are saying i can achieve by manually editing the generated spec file by removing the first line and replacing the key from the second line with the key in the first line as so

<artifact id="Condition/RT-SNF-Diagnosis-CI" key="Condition-BSJ-Diagnosis1" name="RT-SNF-Diagnosis-CI"/>

Seems reasonable, so I have
1. Made the manual edits for this and most other deprecations
2. Committed the change to the IG repo (but not yet the JIRA-Spec-Artifacts repo)
3. Rebuilt the IG locally

Now, based on what you're saying, I had expected the manual changes I made to stick, but the file gets reverted back to the intermediate state with 2 lines. So at this point I think that either
1. I need to commit the changes to the JIRA-Spec-Artifacts repo to make them sticky, or
2. I'm doing something not quite right.
In the case that the answer is 1, I've pushed that update here, but please let me know if I need to adjust still.

Thanks for your patience as I figure out this process

lmckenzi:
Correct.  Commit needs to happen and propagate before it becomes 'sticky'.

karlnaden:
worked like a charm - thank you

PR #596, opened 2022-09-08
Deprecate duplicate CDA eICR specification

minigrrl:


minigrrl:
Lloyd - have moved all tickets that were assigned to these duplicate specifications and now want to deprecate them so that nobody else uses them (we missed some issues because people used the wrong ones)

PR #600, opened 2022-09-13
Add barebones spec file for ICHOM's Breast Cancer IG

vadi2:
Added barebones spec file for ICHOM's Breast Cancer IG - idea being that the IG publisher can flesh out the rest.


lmckenzi (changes requested):
version should be 1.0.0-ballot or current
page should be something like 'home'.  Once a key or version are created, they can't be removed, so we want them to be right.

vadi2:
@lmckenzi I've applied the requested changes.

lmckenzi (changes requested):
Still need to change the 'page'

vadi2:
I already changed it, what do you mean?

lmckenzi:
Ok.  Didn't see it when I looked at changes.  Still need to change default version to 1.0.0

vadi2:
Sure, done.

lmckenzi:
Sorry, I lied.  You've only defined 'current', so the default version needs to be 'current'.

vadi2:
Fixed.

PR #601, opened 2022-09-13
Allow https build.fhir.org links for ciUrl

vadi2:


lmckenzi (commented):
Why make this change?  This will fail all of the existing files that use http:, which is what we're standardizing on.  (My understanding is that with HTTP3, https goes away because everything is implicitly https.)

vadi2:
It will not fail existing files that use http:, it allows for both. See the sample run at https://github.com/vadi2/JIRA-Spec-Artifacts/actions/runs/3046149892/jobs/4908596565 which demonstrates that both http and https work.

vadi2:
I can't say I agree with that reasoning - as of today https: is the gold standard, and many IG's are using such links already.

lmckenzi:
They shouldn't be able to use such links - the validation rules here should prevent that...  Where are you seeing them?

vadi2:
Quite a few in the repo...

image

PR #602, opened 2022-09-14
Rename FHIR-ichom-breastcancer.xml to FHIR-ichom.breastcancer.xml

vadi2:
I missed that the file should be called this. IG QA says:

> There is no jira specification file defined with the expected name of FHIR-ichom.breastcancer in the xml folder of the https://github.com/HL7/Jira-Spec-Artifacts project. This must exist before ballot or publication.


lmckenzi:
Once it exists, you can't rename it.  You'll have to use the jira-code parameter in your IG to point to the code that now exists

vadi2:
Gotcha.

vadi2:
Where can I find documentation for this parameter? It's not in http://build.fhir.org/ig/FHIR/fhir-tools-ig/branches/master/CodeSystem-ig-parameters.html

vadi2:
Found it in https://confluence.hl7.org/pages/viewpage.action?pageId=76153131#ConfiguringSpecificationFeedback-Filetomodify.

PR #605, opened 2022-09-15
Carin stu2

cspears-mitre:
Many of the names of Examples files where renamed and the whole example set was reorganized and a set of FSH and JSON examples were merged. VS and CS have the same name and require renaming in the Jira artifacts names


lmckenzi (commented):
Are the deprecated things removed or just renamed?  If renamed, then the key should be retained and changed to the new id and name

cspears-mitre:
The whole thing is a mess now (I am not the only one that was involved over time or recently). Most of the deprecated things (all of the CodeSystems) are actually deprecated. All of the examples were re-organized by someone else (They were a merger of json examples and FSH examples). I can't honestly say whether the ones marked as deprecated were just renamed or if they are a different example. They do have a different key however. The more this builds the harder it is becoming to maintain.

PR #626, opened 2022-11-08
Added Vulcan Real World Data and Schedule of Activities initial spec files

jduteau:


lmckenzi (commented):
Don't you need to update the Specification file too?

PR #638, opened 2022-11-22
Update ICHOM Breast Cancer IG version to 1.0.0-ballot

vadi2:


vadi2:
@lmckenzi 

> Versions can only be removed if no JIRA tracker references that version.  If this is the case and the version should really be removed, please coordinate with an administrator.

This is indeed the case. I had a `0.1.0-ballot` version previously (per faulty advice from QA page).

vadi2:
@lmckenzi ping

PR #641, opened 2022-11-26
Update to the template for FHIR-pddi

rkboyce:


rkboyce:
Sorry - I thought the issues were all fixed. Will work more on them.

PR #642, opened 2022-11-29
Update FHIR-us-mcc.xml

bheale:
updating version


bheale:
@lmckenzi  ready to merge

lmckenzi:
There's already a 1.0.0 version.  Is 1.0.0 not actually out yet, and you're going to ballot aiming to eventually get to a 1.0.0 release, or are you now working toward a 1.1.0 or 2.0.0 release?
If the former, mark the 1.0.0 version as deprecated (as we don't want anyone reporting issues against it).  If the latter, change to 1.1.0-ballot or 2.0.0-ballot.

bheale:
@lmckenzi I deprecated. this will be the first non-ci build version ever. let me know if it looks good now

PR #646, opened 2022-11-30
Update FHIR-us-davinci-atr.xml

nbashyam:
Update to match build for ballot.


lmckenzi (changes requested):
I think the ballot URL is just going to be 2020Feb, not 2.0.0-ballot...

nbashyam:
> I think the ballot URL is just going to be 2020Feb, not 2.0.0-ballot...

You mean 2023Feb right not 2020Feb ?

lmckenzi:
Yes

nbashyam:
Obsolete.

PR #648, opened 2022-11-30
Create FHIR-uv-emedicinal-product-info.xml

bheale:
This is the initial commit. Done on behalf of Craig Anderson for emedicinal-product-info

### DIFF
--- a/xml/FHIR-uv-emedicinal-product-info.xml
+++ b/xml/FHIR-uv-emedicinal-product-info.xml
@@ -1,0 +1,115 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<specification xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" ballotUrl="http://hl7.org/fhir/uv/emedicinal-product-info/Jan2023" ciUrl="http://build.fhir.org/ig/HL7/emedicinal-product-info" defaultVersion="1.0.0-ballot" defaultWorkgroup="brr" url="http://hl7.org/fhir/uv/emedicinal-product-info">
+<version code="current" url="http://build.fhir.org/ig/HL7/emedicinal-product-info"/>
+<version code="1.0.0-ballot" url="http://hl7.org/fhir/uv/emedicinal-product-info/Jan2023"/>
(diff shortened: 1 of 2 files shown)


bheale:
@lmckenzi epi for Craig Anderson ready for merge - Thanks!

lmckenzi (commented):
Have you updated the specifications file to point to this new spec?

bheale:
nope

bheale:
do you mean regading: SPECS-[family].xml

These files list all of the specifications that exist for a given product family.  There is one file per family.  Pull requests to these files may be submitted by any co-chair, project lead or other individual involved in managing HL7 specifications.  The file structure is as follows:

key: the unique identifier for the specification within the product family (what's actually stored on each issue).  This will not generally be seen by users and should never change once established.  Even if the specification is renamed, the key will remain the same.

name: the descriptive name for the specification.  This is what users will see - and can be freely changed if, for example, the specification is renamed, perhaps due to expanded scope.  It may be helpful to include the old name or abbreviations or other keywords to help users who are searching for the specification when submitting an issue.

deprecated: an optional boolean that can be set for specifications that are no longer actively maintained - and for which HL7 no longer wishes to receive feedback.  This might be specifications that got merged into other specs, specifications that have been retired, etc.  Once a specification is marked as deprecated, it will no longer be possible to choose that specification when submitting new issues or changing issues, but the specification will still appear on historic issues and can be used when s

I was wondering.

lmckenzi:
Yup.  That's the one.  Until you add a row there, your new file will be ignored.

bheale:
cool. I am getting an error when I buildlocal.bat 
     [xslt] ERROR: FHIR specifications that are not deprecated must have a gitUrl attribute that starts with 'https://github.com/HL7/' FHIR-uv-emedicinal-product-info

but I don't see gitUrl described in https://confluence.hl7.org/pages/viewpage.action?pageId=76153131#ConfiguringSpecificationFeedback-\[Family]-[code].xml 

Is there a file in my build I need to modify to have the gitUrl field populated? Or do I just manually add it to what I am creating here?

PR #663, opened 2022-12-05
Update PDDI

brynrhodes:


lmckenzi (commented):
Why are you jumping from 0.2.0 to 2.0.0-ballot?  Shouldn't it be 1.0.0-ballot?

brynrhodes:
Because this is STU2, so trying to keep the version aligned with the STU

lmckenzi:
In your versions, I don't see that there's ever been a 1.0.0 release?  (Or any non-ballot release)

brynrhodes:
Okay, I checked the ballot pool and it is listed as STU1, so I updated the version to 1.0.0-ballot. Once this PR is merged I'll commit the updates to the PDDI repository to change the version and ballot title to match.

PR #675, opened 2022-12-09
2nd time of copying jira-new.xml to JIRASPEC/xml/FHIR-us-.....xml....…

saulakravitz:
…hopefully this is the last time through the loop


saulakravitz:
I'm just regenerating using the jira-new.xml produced by the IG Publisher.

PR #678, opened 2022-12-15
Update SPECS-OTHER.xml

PJHaug:


PJHaug:
I have updated the SPECS-OTHER.xml file and added a OTHER-arden-syntax.xml file to the xml folder.

PR #681, opened 2022-12-30
add FHIR-us-sirb.xml for sIRB IG jira spec file

GitHubUNCStudent:


lmckenzi:
Are all of the artifacts marked as deprecated really deprecated, or are some of them just renamed?  If renamed, don't deprecate, keep the key the same and revise the id and title

lmckenzi:
It looks like you're deprecating and adding when you should just be renaming.  If you've just changed the id and/or title of an artifact that is essentially the same, then keep the key and just change the title and/or id.  Don't deprecate and add.

GitHubUNCStudent:
I don't see an error log this time. Where is it?   

We want these to have different keys because they are different versions (one is before we added the expressions, the drop-downs to drive auto-population, the external data coming from the parent questionnaire while the other key is after all that was added).  None of the examples validate against the old questionnaires, because so much has changed.   Is that OK or is that making this fail?

lmckenzi:
There's no error in the log.  The issue is that the changes being made don't seem to be the right ones.  It appears that you've renamed some of your artifacts.  The tool that generates the "default" updated Jira spec file doesn't understand that, so it assumes you've deleted the old artifacts and added new ones.  If you've renamed, then you should instead keep the old keys and just update the titles and or ids for the artifacts.

PR #683, opened 2023-01-11
Military service history

IoanaSingureanu:


IoanaSingureanu:
The deprecated references were copied over by Sean from mCode. If we could get rid of them IT WOULD AWESOME!

IoanaSingureanu:
> This doesn't make sense:
> 
> * 
> 
> * 
> 
> All versions need to have a version-specific location. You can't just point to the 'base' specification.
> 
> Also, why are all of your artifacts deprecated.


We would like to remove all the deprecated entries and these bullets but I the deprecated entries were copied over form mCode and the bullets were auto generated during the IG publication.

lmckenzi:
Ok.  There's a plan to do that at some point in the future.  In the meantime, can you fix the URL for the 1.0.0 version?

PR #684, opened 2023-01-11
Military service history

IoanaSingureanu:
IG Publication without deprecated references


IoanaSingureanu:
The commit/build seems to fail/timeout. I included the jira spec file below:
[FHIR-us-military-service.zip](https://github.com/HL7/JIRA-Spec-Artifacts/files/11267852/FHIR-us-military-service.zip)

PR #691, opened 2023-01-31
Update CREDS to current structure

DavidPyke:


lmckenzi (commented):
All you've done is add trailing whitespace?  Why?

DavidPyke:
Did more than that, deprecated an artifact and removed pages but I was too dumb to get it right

PR #701, opened 2023-02-21
Changes for HAI LTCF unballoted STU updates

minigrrl:


lmckenzi:
To clarify - none of these versions are deprecated?  Comments can be submitted against any and maintenance will happen against any?

minigrrl:
Currently – yes.

We are discussing versions right now, so that may change in the near future.

Sarah Gaunt (she, her)
Senior Information Analyst
t: +61 410 292 506

PR #711, opened 2023-03-12
Version 3.0.0.  path 2023May

saulakravitz:


lmckenzi (changes requested):
The URLs look wrong.  Should be:
version code="3.0.0" url="http://hl7.org/fhir/us/mcode/STU3"
version code="3.0.0-ballot" deprecated="true" url="http://hl7.org/fhir/us/mcode/2023May"

Also, if you're just going to ballot, 3.0.0 shouldn't be there at all yet.

saulakravitz:
I changed defaultURL to current
Changed 3.0.0 to 2023May....

Is that what needs done?

lmckenzi:
I fixed my comment so it renders properly

saulakravitz:
3.0.0 is gone
3.0.0-ballot is deprecated....

Are we there yet?

lmckenzi:
version code="3.0.0-ballot" url="http://hl7.org/fhir/us/mcode/2023May"

That's what you should have for the ballot.  The new code shouldn't be there at all.

saulakravitz:
Hopefully now we are there.

PR #714, opened 2023-03-15
Update ICHOM for 1.0.0 STU publication

vadi2:


lmckenzi (changes requested):
The URL for 1.0.0 should be ".../STU1", not "/2023Jan"

vadi2:
@lmckenzi adjusted.

PR #720, opened 2023-03-22
DS4P IG: add new version in advance of publication

mojitoj:


lmckenzi (changes requested):
STU1 should have a version of 1.0.0

mojitoj:
thanks @lmckenzi. Hopefully this is good to merge now.

PR #721, opened 2023-03-22
Update CREDS for ballot

DavidPyke:


lmckenzi:
Branch isn't letting me auto-update, so may be a conflict?

lmckenzi (changes requested):
Pretty sure your canonical shouldn't be "fhir-registry-protocols-ig" - what's the approved path in your IG proposal?

The ballot link should end with "2023May

lmckenzi (changes requested):
Based on the IG proposal
url="http://hl7.org/fhir/us/fhir-registry-protocols-ig"
should be
url="http://hl7.org/fhir/us/registry-protocols">

Similarly ballot link should be
http://hl7.org/fhir/us/registry-protocols/2023May

DavidPyke:
Okay, I'll change it

PR #722, opened 2023-03-23
Updating JIRA Spec for fhir.us.ccda aka ccda-on-fhir

jddamore:
Updating based on request from FMG and warning in publisher QA page as part of ballot publication

### DIFF
--- a/xml/FHIR-us-ccda.xml
+++ b/xml/FHIR-us-ccda.xml
@@ -6,6 +6,7 @@
 <version code="1.0.0" url="http://hl7.org/fhir/us/ccda/STU1"/>
 <version code="1.8.0" url="http://hl7.org/fhir/us/ccda/2017Jan"/>
 <version code="1.6.0" deprecated="true" url="http://hl7.org/fhir/us/ccda/2016Sep"/>
+<version code="2.0" deprecated="true"/>
 <version code="1.0" deprecated="true"/>
 <artifactPageExtension value="-definitions"/>
 <artifactPageExtension value="-examples"/>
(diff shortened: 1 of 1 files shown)


jddamore:
<version code="2.0"/> never existed. This was added manually in this github commit 
https://github.com/HL7/JIRA-Spec-Artifacts/commit/d0a280b863d1fa6b086757c8fc6a9f3b684ea128

I have looked in JIRA and no issue for this specification have been logged against 2.0

If you want me to add, I can do so, but it would be confusing since this is version 1.2.0

lmckenzi:
Once a version is committed, it can't be removed.  You can mark it as deprecated.

lmckenzi (commented):
Are you actively maintaining all of the other versions?  At minimum I'd expect 1.8.0 to be deprecated.  If you don't expect to ever make technical corrections to 1.0.0, you could mark it as deprecated too.

jddamore:
We are not maintaining any prior versions at this time. I will mark as deprecated

lmckenzi:
I would have expected 1.1.0 to still be active at least?  It's the most recent 'official' release, so you'd still want people to be able to submit feedback against that.

brettmarquard:
Multi version will be confusing to commenters -- I recommend we take all comments on latest

jddamore:
I agree with @brettmarquard (who was an author on the previous 1.1.0 release). Let's just take all feedback on what's balloted. We acknowledge that there's a lot more work to do on other sections and tying everything together with Composition (i.e. the prior work). Let me know if any issue with merging @lmckenzi

PR #726, opened 2023-03-27
Refresh JIRA Spec (specification line only)

jddamore:
Lynn asked that we adjust our publication-request.status to "ballot" (from "informative") which caused a change in the first line of the JIRA file. Just updating that line.


lmckenzi (commented):
If 1.1.0 is being deprecated, why is it the default?

jddamore:
I don't quite understand how the publisher picks the "default" in the jira-new.xml. This is the line that came from running and doing the diff between jira-new.xml and jira-current.  

I am ok to have "1.2.0-ballot" as the default but also want to make sure I get the CI-build to resolve JIRA-spec warning per the FMG's ask. Not sure of the interplay between "deprecation" on prior versions and "default" in the specification element.

lmckenzi:
If the only thing it's complaining about is that, you can ignore it.  (I'm not completely clear on why the 1.1.0 is deprecated, as typically feedback against that release is allowed until such time as you've got the 1.2.0 release out.)

PR #746, opened 2023-05-10
Add FHIRPath version 2.0

vadi2:
Added FHIRPath version 2.0 so it shows up in the JIRA form. Haven't updated any other artifacts as it doesn't look like the IG publisher build is setup for the [FHIRPath spec repo](https://github.com/HL7/FHIRPath).


vadi2:
Fixed.

PR #748, opened 2023-05-18
Update CDA-ccda-two-one-companion.xml

russellott:
Deprecating Release 3 and Release 4 - Ballot versions. Adding Release 4 version and Release 4 STU Update 1.


russellott:
Good catch - changed the default version as well in new commit.

PR #755, opened 2023-06-12
Updating Cancer-reporting spec (FHIR-us-cancer-reporting)

dadmbc:
Updating Jira Spec post MedMorph RA publication rebuild


lmckenzi (changes requested):
Can you resubmit without the pretty-printing?  It makes comparison easier.

PR #760, opened 2023-06-14
(new) update on FHIR-uv-emedicinal-product-info

gcangioli:


lmckenzi:
What's the rationale for this change?  This isn't going to fix build issues.  "Home" is the normal name we use for the index page.

gcangioli:
Thank you for pointing this out, I supposed that the update was due to the fact a file name change (from home.md to index.md ), I didn't notice that was also renamed also the page title. I reverted that change, so this PR is not needed. Thanks again. :-)

PR #763, opened 2023-07-03
Trying to update the MDI Jira Spec file for 1.1

DianaWright61:
Trying to update the MDI Jira Spec file for 1.1


lmckenzi (changes requested):
The version URLs are still not correct.  Make sure they're correct in the registered package-list.json and in your publication-request.json

DianaWright61:
The version 1.1.0 has not been published; it is going through an STU Update comment period now. That is why the url for 1.1.0 is the ciUrl. Is that the correct approach?   I would like the version 1.1.0 to show in the Jira list so commenters can designate specific resources.

lmckenzi:
You shouldn't define 1.1.0 until such time as you're ready to publish 1.1.0.  You also can't do an STU comment period without doing a frozen snapshot release and formal publication request for it...  When you define a URL for 1.1.0, it will have to be STU1.1.

lmckenzi (changes requested):
The URLs for the versions don't make sense.  1.0.0 should still be /mdi/STU1.  The 1.1.0 URL looks wrong.

Also, why is the table of contents deprecated?

PR #764, opened 2023-07-04
Update artifacts.

minigrrl:


lmckenzi (changes requested):
The default version should be 1.0.0, not 'current'

minigrrl:
Agh, spent so much time agonizing over that decision. 50/50 and of course, I chose wrong!

PR #772, opened 2023-07-23
Add FHIR-v2mappings.xml file.

rhausam:


grahamegrieve (commented):
why did specs-fhir change?

rhausam:
SPECS-FHIR.xml changed because it appeared that I needed to add the 'v2-mappings' key:
<specification key="v2mappings" name="HL7 Version 2 to FHIR"/>
I thought I gathered that from some of the documentation and/or error messages I was trying to resolve - but if I mis-read or mis-applied that, I'm happy to back it out or fix it however it needs to be.
@grahamegrieve

rhausam:
The SPECS-FHIR.xml is clean now. @grahamegrieve

grahamegrieve (approved):
ok

PR #789, opened 2023-08-01
CIMI Labs

nrdavis1:
Adding file for the Laboratory Base Class FHIR Profiles IG


lmckenzi (changes requested):
Initial version should be 1.0.0 ballot.  CI-build URL is wrong, as is ballot URL

PR #800, opened 2023-08-24
Create FHIR-us-pdmp.xml

fmckinney:
New jira spec file for the PDMP on FHIR IG


fmckinney:
Fixed build problems... added gitUrl and CI build URLs to xml file and added us-pdmp to specs-fhir.xml

PR #806, opened 2023-09-14
Update FHIR-us-hsds.xml

SerafinaVersaggi:
Renamed SD for profiles from HSDS to HSD


lmckenzi:
Please submit a new PR with conflicts resolved (if one is needed at all)

PR #807, opened 2023-09-14
Update FHIR-us-hsds.xml

SerafinaVersaggi:
Updated SD for profiles from HSDSprofilename to HSDprofilename; replaced <?xml version="1.0" encoding="UTF-8"?> after removing inadvertently


SerafinaVersaggi:
Replaced <?xml version="1.0" encoding="UTF-8"?> after removing by mistake; updated SD for profiles to rename HSDSprofilename to HSDprofilename

PR #812, opened 2023-09-24
Update Jira Spec for MDI STU 1.1

DianaWright61:


lmckenzi (commented):
There are a whole lot of extra line feeds that make comparison to the original file harder.  Can you commit a version that doesn't have all of that pretty-printing?

DianaWright61:
I think I got rid of the extra line feeds. Hope that works.

PR #813, opened 2023-09-25
Update for STU1.1

minigrrl:


lmckenzi (changes requested):
It looks like a bunch of things have had their ids changed.  This is being interpreted as an add/remove situation.  However, what you should do is simply use the old key with the new id/names.

minigrrl:
> It looks like a bunch of things have had their ids changed. This is being interpreted as an add/remove situation. However, what you should do is simply use the old key with the new id/names.

@lmckenzi does it really matter? This would be a very large time consuming manual job for me to do (unless there is a way to do it that isn't manual??). Like this is just for the jira drop downs, does it really matter if some artifacts have been deprecated when they have been given new ids and new names?

lmckenzi:
It's how we maintain traceability of what change requests have happened to what artifact over time.  It's also used to help verify maturity.  I recognize it's a bit of a pain, but it is something we've expected of other projects.

minigrrl:
Ok

minigrrl:
@lmckenzi so I think there is a problem here. For BFDR (haven't looked at VRCPL yet), most of the deprecated artifacts were deprecated at the previous publication - and the traceability issue wasn't picked up at that point.  

There were 39 deprecated artifacts in 2021 and there are 54 now. There is clearly nothing we can do about the 39 deprecations that have already been published (in fact, if I try to do the thing with the old key/new id, the build crashes).

14 out of those 15 new deprecations are the profiles and examples we deleted. I've fixed the other one.

minigrrl:
BTW, still working on VRCPL

minigrrl:
Ok @lmckenzi VRCPL did have a bunch where we gave them new ids, so I've fixed those up.

lmckenzi (approved):

PR #815, opened 2023-09-27
Update FHIR-us-davinci-dtr.xml

JeffBrown-lantana:


lmckenzi (commented):
Are all of the new 'deprecated' entries true removals or are some of them renames?  If renames, then keep the old key on the new artifact/page

PR #817, opened 2023-09-28
Update for gender harmony version

swmuir:
add version code="1.0.0" url="http://hl7.org/uv/gender-harmony/2023Sep"


jduteau:
This isn't all that is needed to update the JIRA spec file for the new gender harmony version.  I have a different pull request that adds in the proper changes.

PR #828, opened 2023-10-13
Tweaked in the vain attempt to fix the VRDR build

saulakravitz:
Only material changes are deletion of deprecation of some versions.
Everything else is not material, IMHO.


lmckenzi (commented):
Why are you un-deprecating old releases?  You shouldn't need to do that.

saulakravitz:
I'm grasping at straws.

lmckenzi (commented):
Please roll back that part.  Let's see what happens with just the other changes.

PR #838, opened 2023-10-23
New CDA core SD specification

jduteau:
New CDA core SD specification.
Changed the code to not check for OIDs for CDA artifacts since SD agreed that was not needed anymore with the new way of publishing CDA profiles.


lmckenzi (commented):
The CDA management group signed off on removing the OID rule?

jduteau:
https://confluence.hl7.org/display/SD/2023-09-28+Agenda+and+Minutes

PR #843, opened 2023-10-26
CardX HTN ig

nrdavis1:
Addingnew Jira spec file for the CardX HTN IG.


lmckenzi (changes requested):
ci-build url shouldn't include a branch name
ballotUrl should be something like 2023Sep
There are no 0.1.0 releases.  Your first release should be 1.0.0-ballot
You're not adding cardx-htn to the SPECS-FHIR file

Have both of these IGs had their IG proposals approved by FMG and the codes match those approved by FMG?

nrdavis1:
Added row to SPECS-FHIR.xml file.

nrdavis1:
Made changes to the FHIR-cardx-htn.xml file and added a row for cardx-htn to the SPECS-FHIR.xml file.

nrdavis1:
Added git URL to spec file

nrdavis1:
Need to update the CardX spec file to reflect changes made yesterday.  Will close and re-open pull request.

PR #847, opened 2023-11-01
CardX and CIMI Labs updates

nrdavis1:
Changes to CIMI Labs file and CardX file.


lmckenzi (changes requested):
The approved IG code is "hypertension-mgmt", not cardx-htn.
I don't see an IG proposal for CIMI labs at all...

PR #857, opened 2023-11-21
Updated FHIR-us-pacio-adi.xml for STU1 publication

mlterryMitre1:


lmckenzi (changes requested):
The URL for 1.0.0 should be "http://hl7.org/fhir/us/pacio-adi/STU1"

lmckenzi (changes requested):
<version code="1.0.0" url="http://hl7.org/fhir/us/fhir-pacio-adi"/>
should be
<version code="1.0.0" url="http://hl7.org/fhir/us/pacio-adi/STU1"/>

mlterryMitre1:
@lmckenzi - re the error: "_[schemavalidate] /home/runner/work/JIRA-Spec-Artifacts/JIRA-Spec-Artifacts/xml/FHIR-us-pacio-adi.xml:33:182: cvc-identity-constraint.4.1: Duplicate unique value [Care Plan for Unconscious, vegetative state, coma] declared for identity constraint "artifact-name-unique" of element "specification"_.

Since this value is used in two completely different examples, shouldn't this be allowable?
That phrase is used in the examples, one for McBee, the other for Betsy Smith-Johnson.  It is completely reasonable that the same value is used in both locations as they are separate resources for separate patients.

Please advise.

lmckenzi:
Every artifact in the IG must have a unique title.  (When you look at the table of contents, there shouldn't be any duplicate names, because that's confusing to the reader.)

mlterryMitre1:
@lmckenzi - we're going with Dave Hill's PR #859 for the PACIO ADI IG since it passed. I'll remove my PullRequest.

PR #859, opened 2023-11-22
Updated the FHIR-us-pacio-adi.xml file.

dhill:


lmckenzi (commented):
Are the deprecated resources really gone, or is this a situation where ids or names have just changed?

lmckenzi:
Ok, then remove the deprecation lines and put the old keys on the rows with the new names.

PR #862, opened 2023-11-30
Update BFDR and VRCL

saulakravitz:
A lot of the deprecated resources in vr-common have moved to bfdr.
Some have been removed from vr-common because they are duplicates of vrdr content.
A small percentage of the other changes are renaming to make things consistent across the vital records portfolio.
The harmonization/reorganization of content is a prime thrust of the current ballots.


lmckenzi (commented):
Do you want to deprecate 0.1.0 or even 1.0.0?  Would you consider making technical corrections against either release?

There are a *lot* of deprecated artifacts.  Are those true deprecations, or is this a situation where the id and/or name of an existing artifact was changed?  If the latter, then use the old key with the new identifier/name and don't deprecate and add.

lmckenzi:
Poking on my questions

saulakravitz:
I have reviewed.   I think deprecations are real.

PR #867, opened 2023-12-01
Update FHIR-us-eltss.xml

bheale:
Updated the eLTSS IG JIRA Specification file. 
Build was successful locally


bheale:
@lmckenzi eltss JIRA spec ready to be merged!

PR #869, opened 2023-12-04
eLTSS Update change defaultVersion

bheale:
default version parameter correction


bheale:
@lmckenzi I corrected the default version and added a needed additional version deprecation to match what the publisher produces

PR #876, opened 2023-12-07
update bfdr

saulakravitz:


lmckenzi (changes requested):
"joe" is an unusual title...

If the issue is duplicate titles and this is *not* a case where you're just changing the identifier for an artifact, keep the old title and put something like " -old" on the end.  Titles for deprecated artifacts will still show up in any Jira issues that referenced them and in search dropdowns, so they need to remain meaningful

saulakravitz:
I tried without success to recycle the key....then moved to deprecate/replace.

lmckenzi:
That's because you'd already committed with the new key.  Once the new key is there, not much we can do :(

PR #886, opened 2023-12-11
Update _workgroups.xml HL7 Australia Work Groups

brettesler-ext:
Adding HL7 Australia work groups for affiliate balloting


brettesler-ext:
thankyou LLoyd

PR #891, opened 2023-12-12
CRMI IG 1.0.0-ballot2 update

brynrhodes:


lmckenzi (commented):
There are a *lot* of deprecated artifacts here.  Are any of them due to renames, or are they all genuinely gone?

brynrhodes:
Combination of things:

1. There were extension pack extensions that hadn't been published in an extension pack yet last time around, so now that they are in the extension pack I could remove them from the IG
2. Several of the profiles moved to the CQL IG, so those are genuine removals
3. I renamed a bunch of the example artifacts to clarify what examples they were for

lmckenzi:
Renamed ones should keep the old keys rather than deprecating and creating new ones

brynrhodes:
Okay, but then how does the generation know what the key is?

lmckenzi:
If you've changed a name and it's deprecating the old one and adding the new one, grab the old key (deprecated row) and stick it with the new name (new row) and delete the deprecated row.  This has to be done manually.

brynrhodes:
Okay, fixed the ones I could, several of the ones that were changed it complained that they were being removed.

PR #893, opened 2023-12-12
Update version

minigrrl:


lmckenzi (commented):
Do you want the technical correction published at a new location or would you like it to overwrite the original?  The second option is the preferred publication approach unless there's a reason to retain the original.  If that's ok, then change the url for 2.1.1 to be the same as 2.1.0

minigrrl:
Definitely overwrite the original. Thanks! Shall do…

Sarah Gaunt (she, her)
Senior Information Analyst
t: +61 410 292 506

PR #897, opened 2023-12-12
Update FHIR-genomics-reporting.xml

sirkyp:
Removed examples before publication


sirkyp:
@lmckenzi - One more change for genomics-reporting. I tried to make the change we discussed, please double check it :)

PR #905, opened 2023-12-14
Update CARIN Insurance Card

ekivemark:
FHIR-us-carin-dic.xml for 1.1.0 update


lmckenzi (commented):
Should the default version change to 1.1.0?

PR #918, opened 2023-12-27
Update SHC Vaccination & Testing IG XML

masnick:


lmckenzi (commented):
Are any of the deprecated artifacts situations where an artifact has been renamed?

masnick:
> Are any of the deprecated artifacts situations where an artifact has been renamed?

Sorry about that, one of them was. I fixed it in 3211d94.

masnick:
Never mind, it looks like the artifact I thought was renamed was actually already in there, which broke the build. I reverted the change.

PR #924, opened 2024-01-04
Update pdex 2024 01 04

ekivemark:
setting default version to 1.0.0. for fhir-us-davinci-pdex.xml


lmckenzi (commented):
Why isn't the default version 2.0.0?

PR #931, opened 2024-01-18
Updating specification url and ballot version

dadmbc:


lmckenzi (commented):
Should the default not be 1.0.0?

PR #963, opened 2024-03-25
Update FHIR-au-core.xml

brettesler-ext:
adding ballot version


lmckenzi (commented):
Do you really want 0.3.0-ballot?  That means "when I've passed ballot, I will publish as 0.3.0" (which would be unusual to say).  Typically you would have a version of 1.0.0-ballot (meaning this is the first ballot leading toward the eventual publication of 1.0.0.

brettesler-ext:
Hey @lmckenzi we are locked into a different versioning scheme based on history; so can we go for 0.3.0-ballot please.

lmckenzi:
If you update to the current branch, I can merge

PR #970, opened 2024-03-27
Add Deprecated version 3.0.0 for VRDR

saulakravitz:


lmckenzi:
3.0.0 wasn't there before, so don't add it.  There's no reason for this PR.

PR #972, opened 2024-03-28
Update carin dic

ekivemark:
references to ci boule use http


lmckenzi:
The only changes here are whitespace...

PR #981, opened 2024-04-01
update FHIR-CardX-HTN-MNG xml

Callie-Phillips04:


lmckenzi (changes requested):
URL for 1.0.0-ballot should be http://hl7.org/fhir/uv/CardX-HTN-MNG/2024May

PR #997, opened 2024-04-11
Updated CDA Specification - added info for CDA Online Navigation Ballot

LisaRNelson:
Lloyd, I couldn't manipulate the page entries without BuildLocal failures. Maybe you could take a look at why I can't modify the page entries tomorrow.


lmckenzi:
The page keys can't have spaces.

PR #998, opened 2024-04-11
Fixed the page entries for the core spec

LisaRNelson:


lmckenzi (commented):
Why are you deprecating 'many'?  If you're getting rid of the high level categories, you should be listing all of the artifacts individually.  With this change, you can only complain about the home page or links, which doesn't make sense.

PR #1002, opened 2024-04-23
AE clinical res profile R5 and backport update

bheale:
Update for publication AE clinical res profile R5 and backport


lmckenzi (commented):
Do you really want a new URL?  Most of the time with a technical correction, we overwrite the STU1 release

bheale:
That is a good question. I was unsure. I am happy to keep the same URL. that would be best

lmckenzi (changes requested):
I still see
<version code="1.0.1" url="http://hl7.org/fhir/uv/ae-research-backport-ig/STU1.0.1"/>
For the backport IG

bheale:
@lmckenzi all set now, I believe

lmckenzi (approved):

PR #1018, opened 2024-07-01
Create FHIR-uv-smart-health-cards-and-links

fmckinney:


lmckenzi (changes requested):
Current URL can't be empty

fmckinney:
Oops. I added that and the missing git URL into the file

PR #1028, opened 2024-07-18
corrected update for PHD IG 1.1

ErikMoll:


lmckenzi (commented):
Should 1.0.0, 0.3.0 and 0.1.0 all still be 'active'?  Would have thought most of them would now be deprecated (i.e. you're no longer accepting new change requests against those versions)

ErikMoll:
I will make those older versions "deprecated".

ErikMoll:
Marked older versions as deprecated.

PR #1049, opened 2024-08-07
vrcl + deprecating versions that I don't want to get jira tickets about

saulakravitz:


lmckenzi (commented):
Should earlier versions be deprecated?

saulakravitz:
STU1 versions still referenced by active publications.   
Next round of publication we can deprecate 'em.

lmckenzi (commented):
Do you want 1.0.0 and 1.1.0 to still be active?

saulakravitz:
Uncle.  I marked a bunch of versions that I don't want to get jira tickets about as deprecated.

PR #1079, opened 2024-08-28
Updates for vrcl, bfdr, and vrdr publication

saulakravitz:


lmckenzi (changes requested):
In vrdr, the ballotUrl shouldn't be STU3

PR #1080, opened 2024-09-03
CodexRT STu2

saulakravitz:


lmckenzi (commented):
Are you wanting to deprecate getting further feedback against 1.0.0 or 2.0.0-ballot?

saulakravitz:
deprecated old versions, as suggested.

PR #1087, opened 2024-09-12
Updates for new versions of HAI and HAILTC

minigrrl:


lmckenzi (changes requested):
Should any of the versions be deprecated?  Typically when you put out a technical correction you'd deprecate the prior versions because you can't accept changes against the originals anymore.  All changes need to be with respect to the new TC

minigrrl:
@lmckenzi Right now we don't want to deprecate any of the older versions. The latest one won't be started to be implemented for another year. We want to keep them all open for comments.

minigrrl:
Ok, I will update.

Sarah Gaunt (she, her)
Senior Information Analyst
t: +61 410 292 506

minigrrl:
@lmckenzi I have deprecated the versions that can't be updated

PR #1095, opened 2024-10-14
update to cardx-htn-mng-xml

Callie-Phillips04:
Addresses file changes to IG post ballot and resolving the QA report warning


lmckenzi (changes requested):
The ballot URL shouldn't change.
The release URL should be something like .../STU1.2, not .../1.2.0

PR #1098, opened 2024-10-16
Updated FHIR-us-qm.xml

abdullah-git1:


abdullah-git1:
How can I determine the issue if the check indicates that it’s not a valid XML document?

I ran it through an XML validator, and it reports that there are no issues.

lmckenzi:
It's not valid against the schema.  If you look at the details of the run, you'll see this:
[schemavalidate] /home/runner/work/JIRA-Spec-Artifacts/JIRA-Spec-Artifacts/xml/FHIR-us-qm.xml:161:97: cvc-identity-constraint.4.1: Duplicate unique value [Intended Venue] declared for identity constraint "artifact-name-unique" of element "specification".
[schemavalidate] /home/runner/work/JIRA-Spec-Artifacts/JIRA-Spec-Artifacts/xml/FHIR-us-qm.xml:235:109: cvc-identity-constraint.4.1: Duplicate unique value [eCQM Update 2024] declared for identity constraint "artifact-name-unique" of element "specification".
[schemavalidate] /home/runner/work/JIRA-Spec-Artifacts/JIRA-Spec-Artifacts/xml/FHIR-us-qm.xml:236:97: cvc-identity-constraint.4.1: Duplicate unique value [eCQM Update 2024] declared for identity constraint "artifact-name-unique" of element "specification".

abdullah-git1:
Based on those details, for the first bullet point

I’m going to assume that reason its not passing the check because the codesystem and valueset have the same name and should not both be called "intended value"?

<artifact id="ValueSet/intended-venue" key="ValueSet-intended-venue" name="Intended Venue"/>
<artifact id="CodeSystem/intended-venue" key="CodeSystem-intended-venue" name="Intended Venue"/>

lmckenzi:
Right.  Titles have to be unique across all artifacts because they'll all show up as entries in the ToC and multiple titles with the same name will confuse the reader.

PR #1100, opened 2024-10-22
cardx-publication-request-update

Callie-Phillips04:


lmckenzi (commented):
Typically the ballot release would be marked as 'deprecated' once an official release is out (indicating that you don't want any more comments against the ballot release).

Callie-Phillips04:
Thanks Lloyd. Since the official STU1 release is not out yet, should the ballot release be deprecated? We still need FMG to vote on the publication request. Thanks again.

PR #1106, opened 2024-10-29
Update for technical correction

minigrrl:


lmckenzi (commented):
Shouldn't 2.1.1 and 2.1.0 be marked as deprecated?  Presumably you want all feedback against 2.1.2?

minigrrl:
You're correct! Have updated...

PR #1118, opened 2024-11-20
Update FHIR-us-pq-cmc.xml

chn100:
updates to terminologies


chn100:
Thanks,

You did this while I was writing the last message.
________________________________

PR #1156, opened 2024-12-12
Update CCDA-on-FHIR for Jan 2025 Ballot

benjaminflessner:


MLD1995:
Looks like it has a duplicate value.  <page key="FC-medications" name="FHIR to C-CDA Medications"/> appears twice.  Not sure what's up with that.

jlyle1:
removed duplicate page

lmckenzi:
Needs to be updated to the current base branch

lmckenzi:
And again :(  Normally it lets me do it from within Github, but it's not showing that option for some reason.

PR #1157, opened 2024-12-13
Add FHIR-qm.xml

abdullah-git1:


MLD1995:
The build is complaining about the ci build url.  It should be http instead of https.  I believe all that needs to happen is changing that in the publication-request.json file.

abdullah-git1:
@MLD1995 Looks like the Jira Spec warning was resolved in the IG. This PR also contains an incorrect change in the FHIR-uq-cqm.xml. Will close ths PR

PR #1158, opened 2024-12-14
SPHR > PHR Rename

awatson1978:
Per the FHIR Management Group, this implementation guide id is to be `hl7.fhir.uv.phr`, and we are renaming instances from `sphr` to `phr`.


awatson1978:
Replaced with https://github.com/HL7/JIRA-Spec-Artifacts/pull/1159

PR #1186, opened 2025-02-19
Update AU Base and AU Core versions

brettesler-ext:


brettesler-ext:
AU Base and AU Core new versions - ready for merge

PR #1193, opened 2025-03-18
Updates for May 2025 ballot cycle

minigrrl:


lmckenzi (commented):
Why are you not changing the default version if you're incrementing the version?

minigrrl:
@lmckenzi I thought the default version should stay as the published version until the new version is balloted and published. This IG is going to ballot in May so wouldn't be the default version yet right?

lmckenzi (approved):

lmckenzi:
I forgot that CDA still updates versions even when it's just a ballot

PR #1210, opened 2025-03-27
Update after new hl7 base template

mvdzel:


mvdzel:
In the previous version I had the version wrong. It should be 2.1.1-ballot, so I fixed it. But now I get this error :-(

lmckenzi:
You can deprecate a version (and undeprecate it later), but once a version (or anything else) has been committed, it can't be removed.

mvdzel:
So I can solve this by manually adding a 2.1.1 version

mvdzel:
Looks the JIRA specs are correct now.

PR #1221, opened 2025-04-02
Added ehr-data-quality

mvdzel:


lmckenzi (changes requested):
"page 1", "page 2", etc. are not appropriate page names.  The purpose of artifact names is to track feedback against the same things from release to release.  What content appears on one page vs. another will almost certainly be different from release to release. If there aren't section headings to use for breakdown, then just have a single page for 'main content' or something like that.

mvdzel:
The pdf has no chapters, just pages. I have asked Gary.

PR #1230, opened 2025-04-22
Updated EU extensions, Added EU Base IGs (R4 and R5)

gcangioli:
deprecated r5 eu extensions and updated eu extensions


rhausam (approved):
These updates all appear to be appropriate.  The intent is to no longer use "EU Extensions (R5)" (eu-r5-extensions) so that spec key is being deprecated.  Will merge.

PR #1234, opened 2025-04-24
PACIO PFE Jira spec updates for STU-2 ballot reconciliation

bmeshell:


lmckenzi (commented):
Presume the two ballot versions and probably 1.0.0 should be deprecated?  I.e. you're no longer accepting requests to change those?

lmckenzi (commented):
Do you want to deprecate 10.0 (or at least 2.0.0-ballot?

bmeshell:
> Do you want to deprecate 10.0 (or at least 2.0.0-ballot?

Hi Lloyd. Yes, I would like to deprecate 1.0.0, 1.0.0-ballot, and 2.0.0-ballot. 

Can you please give me some guidance on how to do this? When I try to update FHIR-us-pacio-pfe.xml in my project https://github.com/HL7/fhir-pacio-pfe it is overwritten every time I run the IG publisher. 

Which file should I change to indicate these older versions are deprecated? Thank you!

PR #1239, opened 2025-05-01
Updated Jira Spec for STU 2

DianaWright61:


lmckenzi (commented):
To confirm, you *don't* want to deprecate 1.1.0 and will potentially make changes against it (i.e. a 1.1.1 or a 1.2.0 release)?

DianaWright61:
Oh dear, that was NOT my intent. Future changes should be logged and made only against the new STU 2 (2.0.0). Can you adjust that?

PR #1244, opened 2025-05-08
Gravity sdoh 2.3

nussteja:
Jira Spec update for Gravity SDOH Clinical Care IG STU2.3


lmckenzi (commented):
Shouldn't some of the older versions be marked as deprecated?

nussteja:
What would be best practice for that? Deprecate the STU1 versions?

lmckenzi:
Normally you deprecate any versions you don't intend to make changes to going forward.  I.e. if someone submits a technical correction for 1.0.0, would you publish a 1.0.1 or would you address it in a 1.3.1 or 1.4.0?  If the latter, you should deprecate 1.0.0.  (And same logic goes for the other earlier versions.)  Typically we deprecate ballot releases once the ballot is closed too.

nussteja:
Makes sense to me, I have made that change to deprecate STU1 version and merged into my branch

nussteja:
Ah, I misunderstood. No we would not, currently all feedback received would be applied either to a 2.3.1 or a 2.4.0 (though most likely to STU3 which is our current focus). I've deprecated up to STU2.3.

PR #1249, opened 2025-05-21
QICore STU7 Publication Update

brynrhodes:


brynrhodes:
Good question, yes, deprecated versions before 4.1.1,  as well as the 7.0.0-ballot version

lmckenzi (commented):
So 5.0.0 and 6.0.0 are not deprecated?  The work group would consider producing a 5.1.0 or 5.0.1 release if problems were identified (as opposed to addressing them in a post 7.0.0 release)?

PR #1250, opened 2025-05-23
HL7 EU HDR FHIR IG added

gcangioli:
HL7 Europe HDR FHIR IG 0.1.0-ballot


lmckenzi (commented):
0.1.0-ballot means that after ballot you're planning to release 0.1.0?  That's possible, but unusual.  Normally the intent is that post-ballot you'll have 1.0.0, meaning the version should be 1.0.0-ballot.

gcangioli:
This is what we are currently applying https://confluence.hl7.org/spaces/HEU/pages/193659500/Versioning+of+the+HL7+EU+FHIR+IGs
assuming 1.0.0 for the first normative version.

But there are different visions on this so it has not been formally approved yet

gcangioli:
going for 1.0.0 = STU 1 ..2.0.0 = STU2 if totally fine if you plan never have a normative version :-) 

otherwise you may have Edition 1 normative as version 3.x.x and an Edition 2 STU 2 as 5.x.x (still valid, but less clear for end users)

PR #1251, opened 2025-05-27
Updated for DaVinci RA STU2.1

yanheras:
Updated FHIR-us-davinci-ra.xml for davinci risk adjustment STU2.1 publication


lmckenzi (commented):
Don't you want to mark the old versions as deprecated?

I think you're deprecating the wrong thing for coding-gap-remark.  Should be deprecating the value set, no?

Why are you making all of your example names ugly?

PR #1259, opened 2025-06-06
Update FHIR-uv-smart-health-cards-and-links.xml

fmckinney:


lmckenzi (commented):
Do you not want to flag the ballot release as deprecated?  (I.e. you'll no longer accept new change requests against the ballot or publish new releases of the ballot)

fmckinney:
Will resubmit after adding deprecated="true" to the ballot version (thanks Lloyd!)

PR #1268, opened 2025-06-15
Risk Adjustment STU2.1 publication

yanheras:
updated Jira spec. fixed issues with keys.


lmckenzi (commented):
1.0.0 and 2.0.0 aren't deprecated?  I.e. we would consider putting out a technical correction on either of those releases rather than a new update based on 2.1.0?

PR #1272, opened 2025-07-02
Update FHIR-uv-smart-health-cards-and-links.xml

fmckinney:


lmckenzi (changes requested):
This looks like it's a rename.  The key should be kept the same and just change the id and title.

PR #1274, opened 2025-07-04
Unballoted update version update

minigrrl:
CDA IG version number update for unballoted update of HAI LTC AU IG


minigrrl:
This is so that people can make comments on the unballoted update - i.e. 1.1.1 isn't out yet. No other versions to be deprecated until we publish this version.

PR #1279, opened 2025-07-15
Update FHIR-qicore.xml

abdullah-git1:
Updated the FHIR-qicore.xml, not the FHIR-qm.xml


lmckenzi (commented):
Should 7.0.0 or 6.0.0 be deprecated?  I.e. will you be accepting changes against them?

abdullah-git1:
> Should 7.0.0 or 6.0.0 be deprecated? I.e. will you be accepting changes against them?

We won't be deprecating 6.0.0 but will be deprecating 7.0.0

PR #1300, opened 2025-07-26
Sync PHD spec artifacts for upcoming ballot

ErikMoll:


lmckenzi:
Still a question of why you're publishing 2.1, when you balloted as 2.0.

lmckenzi:
I see a 2.0.0-ballot but no 2.0.0 release.  Are you doing another ballot, or are you doing a release?  If you're doing a second ballot, the version should be 2.0.0-ballot2.  If you're doing a release, it should be 2.0.0.  If there's an existing 2.0.0 release, that should be in the package-list for the specification already, but seems not to be, so you'll need to follow up w/ Grahame about what happened.

ErikMoll:
There is no 2.0.0 release yet and indeed we are doing a second ballot. I will check in DEV, but renaming to 2.0.0-ballot2 should be fine.

ErikMoll:
I think this update addresses the outcome of the FMG call yesterday - 2025-07-29.

PR #1329, opened 2025-07-31
Update for ballot (CDA IG)

minigrrl:
Lloyd - this is for a CDA IG that is going to ballot (and for which I have already put in the jira spec file with a new version number (1.4.0) for ballot). TSC has changed the name (and thus also the numbering) and I now want the version number to be 4.0.0.
Assuming this is the correct course of action - deprecating the 1.4.0 with my pull request and that it's too late to just delete the 1.4.0 (there are no comments against it yet)?


lmckenzi:
Right.  Do you want to mark 1.4.0 as deprecated?  (And do you want to leave all the other old versions as not-deprecated?  I.e. you might still publish technical corrections against them?)

minigrrl:
Yes, 1.4.0 deprecated - they others stay not-deprecated (for now, will probably change after ballot). Thanks!

PR #1337, opened 2025-08-26
prepublication of CGM ver 1.0.0

Healthedata1:


Healthedata1:
I just updated with whatever the build creates.  I do not recall ever
manually deprecating a prior release, although I see several instances in
the US Cores Jira file.  What is the trigger in the publisher that causes
the jira-new/xml to be marked as deprecated?

Eric M Haas, DVM, MS
Health eData Inc
1131 Estates Drive, Napa, CA 94558
707.227.2608
***@***.***

lmckenzi:
You flag versions as deprecated manually.  Once it's merged, the 'generated' file will flag the release as deprecated going forward.  So just manually edit your source file to flag the release as deprecated and re-commit.

Healthedata1:
see new pr

PR #1339, opened 2025-08-28
Update FHIR-us-cancer-reporting.xml

Pat-rickJoyce:


lmckenzi (changes requested):
Should any of the older versions be marked as deprecated?  I.e. you're not going to be making new updates against any of them anymore?  (1.0.0 almost certainly should be, given that technical corrections would now need to be based on 1.0.1.)

Pat-rickJoyce:
Hi Lloyd. Can I deprecate by setting both = false within the specification file or is there an additional step? @dadmbc will be helping out as I go out of office shortly.

PR #1474, opened 2026-04-10
Adding Clinical Genomics Incubator to the JIRA Specs

khalifa-aly:
updating the SPECS-FHIR.xml to include the cg-incubator entry and adding the FHIR-cg-incubator.xml file from the IG


lmckenzi (changes requested):
Are you publishing a snapshot?  For what purpose?

If you are, the version should be 1.0.0-snapshot.  Our versions always start with the first version we expect to publish.  We then put a qualifier on the end to indicate this is a pre-release in the lead-up to that eventual publication.

khalifa-aly:
No, just a normal version. I'll change it to current. This would be ok?

lmckenzi (commented):
I still see this: <version code="0.1.0-ci-build" url="http://hl7.org/fhir/uv/cg-incubator/STU1-snapshot"/>

lmckenzi (commented):
I still see this: <version code="0.1.0-ci-build" url="http://hl7.org/fhir/uv/cg-incubator/STU1-snapshot"/>

khalifa-aly:
> I still see this:

Thanks Lloyd, but unfortunately I can't see what you are referring to by "this". It does not show from my end in last two comments.
I assume you mean the snapshot version element. I've removed it. If I need to fix something else, please let me know.
Thanks

lmckenzi:
Looks good now

khalifa-aly:
> Looks good now

Thanks a lot

PR #1478, opened 2026-04-15
EU base and core IG update

gcangioli:
Hi @lmckenzi 
this PR to fix the jira-specs warning in the IG QA; however it is not clear to me if the attributes _ballotUrl_ and _defaultVersion_ in 

<specification xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" 
**ballotUrl="http://hl7.eu/fhir/base-r5/2.0.0**" ciUrl="https://build.fhir.org/ig/hl7-eu/base-r5" **defaultVersion="0.1.0"** defaultWorkgroup="eu" gitUrl="https://github.com/HL7-eu/base-r5" url="http://hl7.eu/fhir/base-r5">

that I fixed by hand in my previous version (ballotUrl="http://hl7.eu/fhir/base-r5/2.0.0-ballot and defaultVersion="2.0.0") are relevant.

In case they are, where they get these data from ?

Thanks,
Giorgio


lmckenzi:
I suspect the information in your publication-request.json is wrong.  I'd suggest updating it, rerunning your build, and see if that fixes the issue.

gcangioli:
Thanks found the error .. the status was still ballot...

PR #1479, opened 2026-04-15
Update for SHC Vaccination IG

masnick:


lmckenzi (changes requested):
0.6.2 should be deprecated as it's a ballot release and you won't be making updates based on it anymore.

masnick:
Thanks @lmckenzi, I added that deprecation.

PR #1489, opened 2026-04-27
EU base and core v2.0 update

gcangioli:
deprecated some of the not used artefacts 
...hoping it is the very last update of those guides


lmckenzi:
In `FHIR-eu-base.xml`, the version `2.0.0-gamma` (pointing at `[http://hl7.eu/fhir/base/xtehr`](http://hl7.eu/fhir/base/xtehr)) is still active and not deprecated, alongside the now-published `2.0.0`. To confirm: do you intend to potentially make a new release based on `2.0.0-gamma` (e.g., a technical correction), or has it been superseded by `2.0.0`? If the latter, it should be marked `deprecated="true"` so people don't submit change requests against it.

gcangioli:
Thanks, deprecated

PR #1490, opened 2026-04-27
Add change log to FHIR OMOP IG

jduteau:


lmckenzi (commented):
The `1.0.0-ballot` version should remain `deprecated="true"`. The official `1.0.0` release has been published (per the `INFORMATIVE1` URL), so the preceding ballot should not be un-deprecated.

If you wouldn't publish a technical correction against `1.0.0-ballot`, it should stay deprecated so people don't submit change requests against it. Could you confirm the intent here — is there a reason to re-activate the ballot?

Also, the PR title is "Add change log to FHIR OMOP IG" but the diff only modifies the deprecation flag on the ballot version; no change-log content appears to be added. Could you clarify whether a separate change is missing from this PR?

jduteau:
Duouh!  Checked in the wrong version!  Updated now.

lmckenzi:
You'll be happy to know that the above analysis of issues came from Claude, and he's now happy with the PR :)

PR #1492, opened 2026-04-29
Update FHIR laboratory specification to version 2.0.0 and deprecate previous versions

patrick-werner:
Preparation vor lab 2.0.0 release

### DIFF
--- a/xml/FHIR-eu-laboratory.xml
+++ b/xml/FHIR-eu-laboratory.xml
@@ -1,125 +1,152 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<specification xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" ballotUrl="http://hl7.eu/fhir/laboratory/0.1.0-ballot" ciUrl="https://build.fhir.org/ig/hl7-eu/laboratory" defaultVersion="0.1.1" defaultWorkgroup="eu" gitUrl="https://github.com/HL7-eu/laboratory" url="http://hl7.eu/fhir/laboratory">
-<version code="current" url="https://build.fhir.org/ig/hl7-eu/laboratory"/>
-<version code="0.1.1" url="http://hl7.eu/fhir/laboratory/0.1.1"/>
-<version code="0.1.0" deprecated="true" url="http://hl7.eu/fhir/laboratory/0.1.0"/>
-<version code="0.1.0-ballot" deprecated="true" url="http://hl7.eu/fhir/laboratory/0.1.0-ballot"/>
-<artifactPageExtension value="-definitions"/>
-<artifactPageExtension value="-examples"/>
-<artifactPageExtension value="-mappings"/>
-<artifact id="StructureDefinition/LabReport" key="StructureDefinition-LabReport" name="A - Laboratory Report"/>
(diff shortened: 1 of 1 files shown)


lmckenzi:
A couple of clarifying questions before this can be approved:

1. **Version sequence — jump from `0.1.1` to `2.0.0`.** The version list goes `0.1.0-ballot` → `0.1.0` → `0.1.1` → `2.0.0`, with no `1.0.0` in between. Per FHIR HL7 conventions, the next official release after the `0.1.x` series would normally be `1.0.0`. Is there a reason this is being published as `2.0.0` rather than `1.0.0`, or was a `1.0.0` (and/or `1.0.0-ballot`) skipped that should be reflected here?

2. **Was there a `2.0.0-ballot`?** I don't see a `2.0.0-ballot` entry in the version list, and `ballotUrl` still points to `[http://hl7.eu/fhir/laboratory/0.1.0-ballot`](http://hl7.eu/fhir/laboratory/0.1.0-ballot). If 2.0.0 went through a ballot cycle, that ballot version should be added to the list and marked `deprecated="true"`, and `ballotUrl` updated accordingly.

3. **Deprecate vs. rename on the artifact list.** This release marks a large number of existing artifacts as `deprecated="true"` (e.g. `StructureDefinition/LabReport`, `StructureDefinition/Subject`, `StructureDefinition/Payer`, `StructureDefinition/Recipient`, `StructureDefinition/Author`, `StructureDefinition/LegalAuthenticator`, `StructureDefinition/Validator`, `StructureDefinition/Order`, `StructureDefinition/Result`, `StructureDefinition/Patient-eu-lab`, `StructureDefinition/Practitioner-eu-lab`, `StructureDefinition/PractitionerRole-eu-lab`, several `*-obl-eu-lab` profiles, `StructureDefinition/HumanName-eu`, `StructureDefinition/Address-eu`, `StructureDefinition/BodyStructure-eu-lab`, etc.) while also adding new artifacts. To confirm: are any of those deprecated artifacts actually being *renamed/replaced* by one of the newly added entries (e.g. `laboratory-accredited-eu`, `Device-measuring-eu-lab`, `Device-specimen-eu-lab`, `animal-specimen-eu-lab`, `specimen-focus`, `Range-eu-lab`, `lab-presenceAbsence-eu-lab`, `lab-medicalDevice-eu-lab`)? If so, those should keep the original `key` and just have the `id`/`name` updated in place rather than deprecate-old + add-new — otherwise the JIRA artifact linkage breaks.

gcangioli:
Hi LLoyd, some answers below

**Version sequence**: the sequence 0.1.x to 2.0.0 is due to the fact that last year the HL7 Europe Technical Coordination Team voted for changing the versioning strategy. This versioning gap can be found also in other HL7 EU FHIR IGs already published

patrick-werner:
Regarding the deprecation vs. rename question:

The listed artifacts were not renamed or replaced by newly added entries. They were marked as deprecated="true" because the corresponding informative obligation profiles and logical models have been removed from this release.

In addition, some profiles have been moved to the base specification. Therefore, there is no direct one-to-one replacement within this IG that would justify keeping the original key and updating the id/name.

The deprecation is intentional and reflects that these artifacts are no longer part of the current publication content.

gcangioli:
**Was there a 2.0.0-ballot**: no this is an unballoted STU update.... it should have been in reality a 1.2.0 version, but for marketing reasons we decided to call it 2.0.

PR #1494, opened 2026-04-29
Update FHIR-us-dapl.xml

nbashyam:
Fix formatting issues


lmckenzi:
A few clarifying questions before this can be approved:

1. The PR deprecates `StructureDefinition/dapl-adverseevent` and adds a new `StructureDefinition/dapl-deidentified-adverseevent`. Is this a rename of the same conceptual profile (in which case the original `key` should be kept and only the `id`/`name` updated in place, to preserve JIRA artifact linkage), or is the original `dapl-adverseevent` a genuinely different artifact that is being retired in favor of a distinct new one?

2. Same question for `StructureDefinition/dapl-allergyintolerance` → `StructureDefinition/dapl-deidentified-allergyintolerance` — rename, or distinct new artifact?

3. Same question for `StructureDefinition/dapl-birthsex-extension` → `StructureDefinition/dapl-sex-extension`. If this is a true conceptual change (birth sex → sex), deprecate+add is correct; if it's a rename, please update the original entry in place instead.

PR #1498, opened 2026-04-29
Update USCore V9 publication

Healthedata1:


lmckenzi:
1. **`ValueSet/us-core-provider-specialty` → `ValueSet/us-core-practitionerrole-specialty`** — the old ValueSet is being deprecated and a new one with a similar purpose is being added in the same PR. Is this a true replacement (different binding/concept), or is it conceptually the same value set being renamed? If it's a rename, please update the `id`/`name` in place on the existing artifact and keep `key="ValueSet-us-core-provider-specialty"` rather than deprecating it and adding a new key.

2. **Library model definitions** — the per-version `Library/uscore-*-model-definition` entries (including `uscore-9.0.0-model-definition`, which is the version this PR is publishing) are all being deprecated, and a new generic `Library/USCore-ModelInfo` is added. Two questions:
  - Is `Library/USCore-ModelInfo` intended to replace the per-version libraries going forward (i.e., a genuine concept change, not a rename)? If so this is fine as-is.

Healthedata1:
>Is Library/USCore-ModelInfo intended to replace the per-version libraries going forward (i.e., a genuine concept change, not a rename)?

Yes  

>ValueSet/us-core-provider-specialty → ValueSet/us-core-practitionerrole-specialty — the old ValueSet is being deprecated and a new one with a similar purpose is being added in the same PR. Is this a true replacement (different binding/concept), or is it conceptually the same value set being renamed? If it's a rename, please update the id/name in place on the existing artifact and keep key="ValueSet-us-core-provider-specialty" rather than deprecating it and adding a new key.

the old ValueSet is being deprecated and a new one with a similar purpose is being added in the same PR.

PR #1502, opened 2026-05-06
HL7 Europe MPD v1.0.0. R4 and R5

gcangioli:
First jira specs version for the HL7 Europe MPD v1.0.0 version R4 and version R5.
deprecated artefacts are logical model no more used


gcangioli:
the ballot release of this ig is the 0.1.0-ballot version. this is due to a new HL7 EU versioning policy agreed after the MPD IG ballot.

PR #1504, opened 2026-05-07
Address FMG feedback including missing example

nbashyam:
Mark several artifacts as deprecated and add new artifacts for de-identified data.

### DIFF
--- a/xml/FHIR-us-dapl.xml
+++ b/xml/FHIR-us-dapl.xml
@@ -6,18 +6,16 @@
 <artifactPageExtension value="-examples"/>
 <artifactPageExtension value="-mappings"/>
 <artifact id="Bundle/dapl-deidentified-bundle" key="Bundle-dapl-deidentified-bundle" name="Bundle Example containing de-identified data using DAPL profiles"/>
-<artifact deprecated="true" id="StructureDefinition/dapl-adverseevent" key="StructureDefinition-dapl-adverseevent" name="DAPL AdverseEvent"/>
+<artifact id="StructureDefinition/dapl-deidentified-adverseevent" key="StructureDefinition-dapl-adverseevent" name="DAPL De-identified AdverseEvent"/>
 <artifact id="StructureDefinition/dapl-age-extension" key="StructureDefinition-dapl-age-extension" name="DAPL Age Extension"/>
-<artifact deprecated="true" id="StructureDefinition/dapl-allergyintolerance" key="StructureDefinition-dapl-allergyintolerance" name="DAPL AllergyIntolerance"/>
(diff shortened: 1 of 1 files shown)


lmckenzi:
A few of the deprecate-and-add patterns in this diff look like they could be renames rather than genuinely new artifacts. Could you confirm the intent for each?

1. `ValueSet/dapl-birthsex-codes` → `ValueSet/dapl-sex-codes` and `StructureDefinition/dapl-birthsex-extension` → `StructureDefinition/dapl-sex-extension` — these look like terminology renames (birth sex → sex). If they are conceptually the same ValueSet/Extension just renamed, the original `key` should be retained and the `id`/`name` updated in place rather than deprecating the old entry and adding a new one (this preserves JIRA artifact linkage). If these are genuinely distinct concepts (e.g., a semantic shift from "birth sex" to "current sex"), then deprecate+add is correct.

2. `StructureDefinition/dapl-adverseevent` → `StructureDefinition/dapl-deidentified-adverseevent` and `StructureDefinition/dapl-allergyintolerance` → `StructureDefinition/dapl-deidentified-allergyintolerance` — are the original (non-de-identified) profiles genuinely being replaced by the new de-identified variants (i.e., the originals are gone), or are these renames? If the originals are simply being repurposed/renamed as their de-identified counterparts, please update the `id`/`name` in place under the original `key`.

nbashyam:
They are renames Lloyd based on feedback that was received. For e.g, Birth Sex is no longer promoted in favor of sex. 
There was one example which was missed in previous checkins which was added.

nbashyam:
@Lloyd, do you need anything else from me ?

lmckenzi:
Thanks for confirming these are renames. Since they're conceptually the same artifacts, the current deprecate-old + add-new approach will break JIRA artifact linkage. Please update the four entries in place — keep the original `key`, just change the `id` and `name`:

1. `ValueSet-dapl-birthsex-codes` — remove `deprecated="true"` and the separate `ValueSet-dapl-sex-codes` entry; on the original line update `id` to `ValueSet/dapl-sex-codes` and `name` to `DAPL Sex Codes` (key stays `ValueSet-dapl-birthsex-codes`).

2. `StructureDefinition-dapl-birthsex-extension` — same treatment: drop the new `StructureDefinition-dapl-sex-extension` entry, un-deprecate the original, and update its `id`/`name` to the sex-extension values.

3. `StructureDefinition-dapl-adverseevent` — drop the new `StructureDefinition-dapl-deidentified-adverseevent` entry, un-deprecate the original, and update its `id` to `StructureDefinition/dapl-deidentified-adverseevent` and `name` to `DAPL De-identified AdverseEvent`.

4. `StructureDefinition-dapl-allergyintolerance` — same treatment: update `id` to `StructureDefinition/dapl-deidentified-allergyintolerance` and `name` to `DAPL De-identified AllergyIntolerance` under the original key.

nbashyam:
@lmckenzi , I have done the above changes.

PR #1505, opened 2026-05-08
Update for unballoted update

minigrrl:


lmckenzi (changes requested):
A couple of items to address before this can merge:

1. **`defaultVersion` should point to `4.4.1`, not `4.4.0`.** This PR adds `4.4.1` as a new official release, so the default should be updated to match the most recently published official release.

2. **`4.4.0` should be marked `deprecated="true"`** Now that `4.4.1` has been published under the same minor release, any technical corrections would be issued against `4.4.1`

minigrrl:
Sorry, @lmckenzi, I should have been more clear. This is for a CDA IG unballoted update that is going through review (i.e.. I need the 4.4.1 to show up in Jira so that people can make comments against that version of the spec). It hasn't been published yet, so 4.4.0 is still the latest version.

minigrrl:
Though, thinking about this - maybe it should be 4.1.1-update? And maybe we also need to add "-update" (or similar) to our FHIR versioning/specification paths here: https://confluence.hl7.org/spaces/FMG/pages/453903648/Specification+Versioning+IG+Publication+Paths?

minigrrl:
@lmckenzi what are your thoughts on this?

lmckenzi:
Sorry, missed the question.  Why are you doing a public review for a technical correction?  Those normally just go through Dan.  If it's not a technical correction, then this should be 4.1.0-snapshot

minigrrl:
It's not a technical correction, it's an unballoted STU update. It's following the number schema that has been used for CDA HAI (Release 4, STU 4, Update 1) = 4.4.1 (previous was a balloted STU at 4.4). The numbering was approved by CDA Mgmt Group, but maybe we can correct it, if you think it's wrong, when it gets published?

lmckenzi:
At some point, I think the TSC is going to force everyone to use semver, but I guess not yet.  It should still be 4.4.1-SNAPSHOT though, if this is the preview version leading up to the 4.4.1 release?  Or is this the actual publication of 4.4.1 after the review is already done?

minigrrl:
Thanks @lmckenzi , it's the preview version - have updated to -snapshot.

PR #1509, opened 2026-05-13
updating FHIR Clinical Document spec for upcoming 1.1 STU update

russellott:


lmckenzi (changes requested):
default version should change to 1.1.0?  
Should earlier versions be deprecated?

russellott:
> default version should change to 1.1.0? Should earlier versions be deprecated?

No - not yet.  We want 1.1.0 available so we can set the "fixed in version" field as we apply changes before publication, and want to keep 1.0.1 as the default and active until we actually publish.

lmckenzi:
Are you not publishing 1.1.0 imminently?  Is this a ballot or snapshot publication?  If you're prepping for an official release, the expectation is that the default gets updated now, even if it might take a few days to a week before publication actually occurs.

jddamore:
@lmckenzi I think we can close this branch and NOT merge any changes. I was looking around and then noticed all the requested changes to default version and the deprecation of prior versions were addressed in this PR: https://github.com/HL7/JIRA-Spec-Artifacts/pull/1531

PR #1510, opened 2026-05-15
updates for new release

grahamegrieve:


lmckenzi:
The ballot and tc1 releases should be marked as deprecated

PR #1517, opened 2026-05-28
Add FHIR-ebm-incubator.xml

brianalperMD:


lmckenzi (changes requested):
The `1.0.0-ballot` version is marked `deprecated="true"`, but it is the only release for this new IG — there is no corresponding official `1.0.0` release yet. A ballot should not be deprecated before its successor official release exists, otherwise there is no active release for people to submit feedback or change requests against.

To confirm: was the `deprecated="true"` on `1.0.0-ballot` intentional? If `1.0.0` has not yet been published, please remove `deprecated="true"` from the `1.0.0-ballot` entry:

lmckenzi (changes requested):
1. The `ballotUrl` on the `<specification>` element is `[http://hl7.org/fhir/uv/ebm-incubator/2026May`](http://hl7.org/fhir/uv/ebm-incubator/2026May), but the `<version code="1.0.0-ballot">` URL is `[http://hl7.org/fhir/uv/ebm-incubator/2026September`](http://hl7.org/fhir/uv/ebm-incubator/2026September). Which ballot cycle is this for? The two URLs should point at the same season.

2. The HL7 ballot season segment uses the three-letter abbreviation rather than the full month name. If this is the September 2026 ballot, the segment should be `2026Sep` (so `[http://hl7.org/fhir/uv/ebm-incubator/2026Sep`](http://hl7.org/fhir/uv/ebm-incubator/2026Sep)), not `2026September`.

lmckenzi:
The `gitUrl` is `[https://github.com/HL7/ebm`](https://github.com/HL7/ebm) but the `ciUrl` is `[https://build.fhir.org/ig/HL7/ebm-incubator`](https://build.fhir.org/ig/HL7/ebm-incubator). These normally point at the same repo. To confirm: does the source for the ebm-incubator IG actually live in the `HL7/ebm` repository, or should `gitUrl` be `[https://github.com/HL7/ebm-incubator`](https://github.com/HL7/ebm-incubator) to match the CI build path?

PR #1519, opened 2026-06-02
CRMI STU2 Publication

brynrhodes:


lmckenzi:
One question:
**Parallel `crmi-*` vs non-prefixed StructureDefinitions** — STU2 adds a set of new StructureDefinitions without the `crmi-` prefix that mirror existing ones, with both kept active:

   | New entry | Existing entry (not deprecated) |
   |---|---|
   | `StructureDefinition/publishablecodesystem` ("Publishable CodeSystem") | `StructureDefinition/crmi-publishablecodesystem` ("CRMI Publishable CodeSystem") |
   | `StructureDefinition/publishableconceptmap` | `StructureDefinition/crmi-publishableconceptmap` |
   | `StructureDefinition/publishablenamingsystem` | `StructureDefinition/crmi-publishablenamingsystem` |
   | `StructureDefinition/publishablevalueset` | `StructureDefinition/crmi-publishablevalueset` |
   | `StructureDefinition/shareablecodesystem` | `StructureDefinition/crmi-shareablecodesystem` |
   | `StructureDefinition/shareableconceptmap` | `StructureDefinition/crmi-shareableconceptmap` |
   | `StructureDefinition/shareablenamingsystem` | `StructureDefinition/crmi-shareablenamingsystem` |
   | `StructureDefinition/shareablevalueset` | `StructureDefinition/crmi-shareablevalueset` |
   | `StructureDefinition/computablevalueset` | `StructureDefinition/crmi-computablevalueset` |
   | `StructureDefinition/expandedvalueset` | `StructureDefinition/crmi-expandedvalueset` |

   Are these intended as truly distinct profiles that coexist in STU2, or are the non-prefixed entries replacements for the `crmi-*` ones? If they're conceptually the same artifact (just renamed by dropping the prefix), the original `crmi-*` keys should be preserved by updating id/name in place — adding new keys breaks JIRA artifact linkage for any prior tickets filed against the original artifact. If they're separate profiles, no change needed.

brynrhodes:
We have opted to keep them as distinct profiles for now to get experience with having a profile in an IG that doesn't have the same canonical base as the IG. I'm not 100% convinced it will work downstream and so I wanted to preserve the option that we know works until we're sure that there aren't any gotchas with using the fhir-based URLs. The ones without the crmi- prefix are actually fhir-based URLs (though that doesn't show up in the JIRA spec).

PR #1520, opened 2026-06-03
Update to help troubleshoot issue with CRMI

brynrhodes:
This is what the build generates, it has a duplicate key, "Parameters/manifest", appearing on both:

```xml
<artifact deprecated="true" id="Parameters/manifest" key="StructureDefinition-cqf-expansionParameters" name="Expansion Parameters"/>
```

and

```xml
<artifact deprecated="true" id="Parameters/manifest" key="Parameters-manifest" name="Input Expansion Parameters"/>
```

I tried renaming the key for the first one to `StructureDefinition/cqf-expansionParameters`, which seems right, and I don't know how that extension definition got the key "Parameters\manifest", but it didn't take and the generation still builds it that way, so I still get the warning.


lmckenzi:
This is what's in the JiraSpec now:
```<artifact deprecated="true" id="StructureDefinition/cqf-expansionParameters" key="StructureDefinition-cqf-expansionParameters" name="Expansion Parameters"/>```

Try manually editing to be that key

brynrhodes:
So that works to fix the build, but then it doesn't represent any actual changes from what is there now, so I'm thinking the issue will persist, but happy to try it after this gets merged?

lmckenzi:
There's nothing here to merge.  You're saying that when you run locally, there's no JiraSpec issue but there is when you run on the CI-build?

brynrhodes:
When I run the publisher, it generates a JIRA spec file that has an error in it, the duplicate key, and so I get the JIRA spec warning in the IG build, but I obviously can't commit the "updated" JIRA spec file because it has the error in it. If I fix the error, as you point out, there's nothing to commit, but then why does the IG build generate the incorrect file?

PR #1523, opened 2026-06-05
Update defaultVersion in XML specification

mvdzel:
Updated defaultVersion in specification element from '2.1.1-ballot' to '2.1.1'.


lmckenzi (changes requested):
While updating the `defaultVersion`, the `2.1.1-ballot` version entry should also be marked `deprecated="true"` now that the official `2.1.1` release is published:

```xml
<version code="2.1.1-ballot" url="http://hl7.org/ehrs/uv/ehrsfmr2/2025May" deprecated="true"/>
```

PR #1537, opened 2026-07-16
Update defaultVersion in XML specification

mvdzel:


lmckenzi (approved):
Now that the official `2.0.0` release is being added, the `2.0.0-ballot` version should be marked `deprecated="true"`

mvdzel:
This should be it.

PR #1539, opened 2026-07-18
Add US Quality Core to Jira

Healthedata1:


lmckenzi (changes requested):
The `ballotUrl` and the `1.0.0-ballot` version URL both point to `[http://hl7.org/fhir/us/quality-core/STU1`](http://hl7.org/fhir/us/quality-core/STU1). Ballot URLs should use the HL7 ballot season segment (e.g., `.../2026May` or `.../2026Sep`), not `STU1` — `STU1` is the path pattern for a published STU release, not a ballot snapshot. Could you update the ballot URL to the correct ballot season?

lmckenzi (commented):
The URL for the ballot release should also be the ballot URL
`<version code="1.0.0-ballot" url="http://hl7.org/fhir/us/quality-core/STU1"/>`
Make sure things are right in your publication-request.json

lmckenzi (approved):

PR #1540, opened 2026-07-21
Update artifacts file for the PoCD IG before ballot start

ErikMoll:


lmckenzi (changes requested):
1. **`defaultVersion` should be updated to `1.0.0`.** With the STU1 release being added, `defaultVersion` should point to it rather than the `0.3.0` ballot.

2. **The `0.3.0` ballot should be marked `deprecated="true"`.** Its URL (`.../2021Sep`) matches the `ballotUrl`, so it's the ballot for the STU1 release being added here. Once STU1 is published, you would not issue a technical correction against the ballot, so it should be deprecated so people don't file change requests against it. To confirm — you don't intend to publish a technical correction against `0.3.0`, correct?

grahamegrieve:
1. **Ballot URL** — `ballotUrl` and the `1.0.0-ballot3` version URL are both `[http://hl7.org/fhir/uv/pocd/STU2`](http://hl7.org/fhir/uv/pocd/STU2%60). Ballot URLs should use an HL7 season segment (e.g., `.../2026Sep`), not the `STU2` path. `STU2` looks like the eventual official-release URL rather than a ballot URL. What is the correct season for this ballot?

2. **Version numbering** — The version list jumps from `0.3.0` to `1.0.0-ballot3`, with no `1.0.0-ballot` or `1.0.0-ballot2` present. Were there earlier 1.0.0 ballots? If so, they should be listed (as deprecated). If not, this should be `1.0.0-ballot` rather than `-ballot3`.

3. **Deprecating `0.3.0`** — `0.3.0` was previously the `defaultVersion` (no `-ballot` suffix) and is now marked `deprecated="true"` while `defaultVersion` moves to a ballot. To confirm: does the WG intend that no further technical corrections will be issued against `0.3.0`, and that all future work is in the 1.0.0 line? If `0.3.0` was the last official release, `defaultVersion` should generally continue to point at it (not at a ballot) until an official `1.0.0` publishes.

ErikMoll:
@grahamegrieve :
1. I updated the version to 1.0.0-ballot. All previous versions, including 0.3.0, were in fact ballot versions. There was no official publication done.
2. I updated the ballot URL to http://hl7.org/fhir/uv/pocd/2026sep
3. The workgroup does not plan further work on the older versions. Only on the 1.0.0-ballot version. Ideally that will lead to a 1.0.0 publication.

Can you merge the PR?

PR #1541, opened 2026-07-21
update for Bulk Sept Ballot

Healthedata1:


lmckenzi (changes requested):
The URL doesn't match ballot URL conventions - do you have permission from Lynn for it to be different?  Normally just .../2026Sep

Healthedata1:
On today's FMG IG Ballot review call today we were presented with this:

[image: image.png]

Which I was told you have a hand in creating.  So we updated our
publication-request.json spec accordingly and the Jira spits it out.


Eric M Haas, DVM, MS
Health eData Inc
1131 Estates Drive, Napa, CA 94558
707.227.2608
***@***.***

lmckenzi:
I create it based on what's in the publication-request.json.  If that's wrong, the proposed JiraSpec will be wrong.

Healthedata1:
This is the path we were told to use:
image

grahamegrieve:
The only change since the prior review is a merge from master (workflow file only) — the outstanding ballot URL question hasn't been addressed.

The `ballotUrl` is still `[http://hl7.org/fhir/uv/bulkdata/4.0.0-202609-ballot`](http://hl7.org/fhir/uv/bulkdata/4.0.0-202609-ballot%60). The HL7 convention is a season segment like `.../2026Sep`, not `202609`. As Lloyd noted, the JiraSpec URL is derived from `publication-request.json` — so if that file uses `4.0.0-202609-ballot`, that's what needs to be corrected first, and then this file regenerated.

To confirm: did Lynn explicitly approve the `4.0.0-202609-ballot` path as an exception to the normal `2026Sep` convention? If not, please update `publication-request.json` in the bulk-data repo to use the standard `2026Sep` segment and regenerate this file. If Lynn did approve the exception, could you point to that confirmation so we can proceed?

PR #1543, opened 2026-07-23
Deprecate snapshot, add new release number

minigrrl:
Note: NHSN is using all the other releases - they are not deprecating any others at this time.


grahamegrieve:
Now that 4.4.1 is published as a technical correction under the 4.4.x line, will you potentially issue any further technical corrections against 4.4.0, or will all future fixes be in 4.4.1 and later? If the latter, 4.4.0 should be marked `deprecated="true"` so change requests are not filed against it.

PR #1545, opened 2026-07-28
Gravity SDOH Clinical Care STU3 Publication

LatashaEstrada:


grahamegrieve:
Now that 3.0.0 (STU3) is the published release and the new default, is 2.3.0 (STU2.3) intended to continue receiving technical corrections? All the earlier 2.x releases are marked `deprecated="true"`, so leaving 2.3.0 active reads as "still maintained." If the workgroup wouldn't publish a technical correction against 2.3.0, it should be deprecated too so that people don't submit change requests against it.

LatashaEstrada:
Thank you Graham for your note on the 2.3 version. Please find an updated commit with the 2.3 version deprecated.

lmckenzi (approved):

PR #1551, opened 2026-08-05
Update PIQI for 1.0 pub and 2.0 ballot

jddamore:
This commit includes both the 1.0.0 publication and the 2.0.0-ballot (September 2026) for PIQI


lmckenzi (changes requested):
1. **`defaultVersion` should be `1.0.0`, not `2.0.0-ballot`.** Now that 1.0.0 is an official release, the default should point at it rather than at the new ballot. The ballot should remain listed as a version, but `defaultVersion` needs to stay on the most recent official publication.

2. **Ballot URL segment.** The new ballot URL should be `http://hl7.org/xprod/ig/uv/piqi/2026Sep`

jddamore:
@lmckenzi  I've made those requested changes. I also noticed that there's a OTHER-piqi-framework.xml in the main JIRA-Spec directory (not the xml directory). I've removed that since I'm assuming it's not needed (also I got confused when I accidentally opened)

jddamore:
I have reverted the ballotUrl change per email from Lynn Laasko and new policy: https://confluence.hl7.org/spaces/FMG/pages/441128396/Proposed+Specification+Versioning+IG+Publication+Paths

That said, the change to the defaultVersion is left per @lmckenzi correction.

lmckenzi (approved):

PR #1552, opened 2026-08-07
Adding SQL-on-FHIR v3.0.0 ballot version of the XML schema.

GinoCanessa:


grahamegrieve:
Claude notes:

A couple of questions on the version list before this can be approved:

1. The `<version>` entries in this file are only `current` and `3.0.0-ballot`, but the version number jumping straight to `3.0.0-ballot` implies that `1.x` and/or `2.x` releases of SQL on FHIR already exist. If there are prior published (STU) releases, could they be added to this file as well? For reference, `FHIR-us-pacio-pfe.xml` (also at the `3.0.0-202609-ballot` cycle) enumerates its earlier `1.0.0`, `1.0.0-ballot`, `2.0.0-ballot`, etc. entries — this one should probably follow the same pattern.

2. `defaultVersion="3.0.0-ballot"` — if a prior official release (e.g., `2.0.0`) has been published, `defaultVersion` should point to that most-recent official release rather than to the current ballot. Could you confirm whether an official release exists, and if so update `defaultVersion` accordingly?

If there truly is no prior official published release of SQL on FHIR and the numbering was chosen for other reasons, a short note to that effect would be helpful — in that case having the ballot as `defaultVersion` is fine.

grahamegrieve:
In fact, there are reasons, so approved

PR #1553, opened 2026-08-07
Update FHIR-ebm-incubator.xml

brianalperMD:
Correcting the ballotURL in the JIRA Spec Artifact for EBM Incubator IG -- does anything else need to be done with the JIRA Spec Artifact?


brianalperMD:
Closing becuase we are doing a snapshot instead of a ballot at this time

PR #1554, opened 2026-08-07
US CARIN DIC: add 2.0.0-ballot version and sync artifacts/pages for STU 2 ballot

ekivemark:
Aligns `xml/FHIR-us-carin-dic.xml` with the current IG content ahead of the CARIN Digital Insurance Card **2.0.0-ballot** (STU 2 Ballot) publication. The IG's `qa.html` has been flagging this file as out of date; this PR clears that warning.

The spec file content is taken from the IG Publisher's proposed `template/jira-new.xml`, reindented to match this file's existing 4-space formatting so the diff stays readable. It validates against `schemas/specification.xsd`.

`us-carin-dic` is already registered in `xml/SPECS-FHIR.xml`, so no change is needed there.

### Versions
- Added `2.0.0-ballot` → `http://hl7.org/fhir/us/insurance-card/2.0.0-202609-ballot`
- Updated `ballotUrl` to the same `2.0.0-202609-ballot` path (per the new FMG path policy for this cycle)
- Marked `0.1.0` (the 2022Jan STU 1 ballot) as `deprecated`
- `defaultVersion` is left at `1.1.0` — this is what the IG Publisher generates, since 2.0.0 has not been published. Happy to change it to `2.0.0-ballot` if that is preferred for ballot comment intake; I kept the generated value so the IG's QA check stays consistent.

### Artifacts added (new since STU 1.1)
- `StructureDefinition/benefit-limitation` — Benefit Limitation Extension
- `StructureDefinition/excluded-services` — Excluded Services Extension
- `StructureDefinition/sbc-insurance-plan`, `StructureDefinition/sbc-metadata`
- `CodeSystem/sbc-benefit-category`, `ValueSet/sbc-benefit-category`
- `CodeSystem/sbc-plan-type`, `ValueSet/sbc-plan-type`
- Examples: `Organization/ExampleIssuerOrg`, `InsurancePlan/SBCExampleHMO`

### Pages added
Use case pages (Authorization and Sharing, Distribution, Patient Access API, Multiple Coverages, Multiple Cards, Consumer Access, Consumer Access and Exchange, RTPBC, IPS comparison, Summary of Benefits), Best Practices, Narrative Conformance, and the SBC pages (sbc, sbc-mapping, sbc-profiles, sbc-terminology, sbc-examples).

### Renamed
- `Use_Case` page name: "Use Case" → "Use Cases"

No existing keys have been removed or changed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


lmckenzi:
Version `1.0.0` (URL `/STU1`) is left un-deprecated alongside `1.1.0` (URL `/STU1.1`). Given that `1.1.0` looks like an update within the STU1 line, would you potentially publish a technical correction against `1.0.0`, or should it be marked `deprecated="true"` as well so that change requests are directed against `1.1.0` (and eventually the STU2 line)?

ekivemark:
Good catch - `1.0.0` should be deprecated. Done in the latest commit.

`1.1.0` is an update within the STU 1 line and supersedes `1.0.0`, and the STU 2 line is now going to ballot. There is no technical correction planned against `1.0.0`, so change requests should be directed at `1.1.0` or the STU 2 line. This matches the treatment already applied to Da Vinci PDex, where `1.0.0` is likewise deprecated upstream.

`1.1.0` is left active, since it remains the current published STU 1 release.

One related observation, in case it is useful. The IG Publisher generates `1.0.0` *without* the `deprecated` flag, so this stays a hand-maintained difference between the file here and the publisher's proposed `jira-new.xml`. That appears to be harmless - PDex carries the same difference and its CI build reports 0 errors / 0 warnings - so the comparison evidently ignores the attribute.

The likely reason the publisher does not derive it is that the published `package-list.json` for this IG has `1.0.0` and `0.1.0` in sequence `"STU 1"` (with a space) while `1.1.0` is in `"STU1"` (no space), so the STU 1 line is split across two sequence names and `1.1.0` is not seen as superseding `1.0.0`. We are following that up separately with the publishing team; it does not affect this PR.

ekivemark:
Lloyd , I can set 1.0.0 to deprecated.

PR #1555, opened 2026-08-07
Using CQL STU 3 Ballot

brynrhodes:


lmckenzi (changes requested):
1. **Ballot URL segment.** The new `ballotUrl` is `[http://hl7.org/fhir/uv/cql/3.0.0-202609-ballot`](http://hl7.org/fhir/uv/cql/3.0.0-202609-ballot). The HL7 ballot URL convention uses a season segment (e.g., `2026Sep`), which is also what the prior ballot entries in this same file use (`2.0.0-ballot` → `2025Jan`, `1.0.0-ballot` → `2024Jan`). Should this be `[http://hl7.org/fhir/uv/cql/2026Sep`](http://hl7.org/fhir/uv/cql/2026Sep) on both the `ballotUrl` attribute and the `3.0.0-ballot` `<version>` URL?

2. **`StructureDefinition/cql-evaluationresult` → `cql-evaluatedResource`.** The diff deprecates `StructureDefinition/cql-evaluationresult` ("CQL Evaluation Result") and adds `StructureDefinition/cql-evaluatedResource` ("CQL Evaluated Resource"). The IDs are similar enough that this could be a rename rather than a genuinely new artifact. Is the new profile a distinct concept, or should the old entry have been updated in place (keep the `key`, change the `id`/`name`) to preserve JIRA artifact linkage?

brynrhodes:
1. Verified the ballot URL is correct, the convention has changed per FMG policy.

2. cql-evaluatedResult is a Parameters profile and the name was changed to cql-parameters because it more accurately reflects the usage (it's not just for output, could be input as well). cql-evaluatedResource is an extension definition and is a totally different thing. So no, this is not a missed rename.

lmckenzi (commented):
Your explanation — *"the name was changed to `cql-parameters` because it more accurately reflects the usage"* — describes a rename of the same conceptual artifact (a Parameters profile whose name was updated to better fit its role). Per the repo convention, a rename should keep the original `key` and update the `id`/`name` in place, so that JIRA artifact linkage survives across the rename.

You already applied that pattern correctly to the example instance:

```xml
<artifact id="Parameters/cql-parameters-example" key="Parameters-cql-evaluationresult-example" name="Example Evaluation Result Parameters"/>
```

(id changed, key preserved).

But the StructureDefinition was implemented as deprecate + add with a different key:

```xml
<artifact deprecated="true" id="StructureDefinition/cql-evaluationresult" key="StructureDefinition-cql-evaluationresult" name="CQL Evaluation Result"/>
...
<artifact id="StructureDefinition/cql-parameters" key="StructureDefinition-cql-parameters" name="CQL Parameters"/>
```

Could you collapse those into a single entry that keeps the original key, i.e.:

```xml
<artifact id="StructureDefinition/cql-parameters" key="StructureDefinition-cql-evaluationresult" name="CQL Parameters"/>
```

The separate `StructureDefinition/cql-evaluatedResource` extension entry is fine to keep as a genuinely new artifact — the concern is only the evaluationresult → parameters rename.

Separately, could you confirm the intent behind newly deprecating `Parameters/qicore-modelinfo-settings` and `Parameters/uscore-modelinfo-settings`? Are those being retired in favor of the generic `Parameters/fhir-modelinfo-settings`, or is a replacement expected in a later commit?

brynrhodes:
And yes, the qicore and uscore modelinfo settings files were deprecated because they required the US Realm dependencies which were removed.

brynrhodes:
Okay, fixed now, sorry for the churn

lmckenzi (approved):

PR #1557, opened 2026-08-08
US Quality Core STU1 ballot

yanheras:


lmckenzi:
For the Task examples, could you clarify the intent?

- `Task/rejected-with-code-example` and `Task/rejected-example` are being deprecated (their names are retained).
- At the same time, `Task/negation-with-code-example` and `Task/negation-example` are being renamed to names that are effectively identical to the deprecated ones (only spacing differs — e.g. `TaskRejected example rejecting a proposal using a code` vs `Task Rejected example rejecting a proposal using a code`).

A couple of possibilities:

1. These really are four distinct examples in the IG, and the `rejected-*` pair is being retired while the `negation-*` pair is coincidentally being renamed. In that case, having two artifacts (one deprecated, one active) carry essentially the same name may confuse consumers — worth differentiating the surviving names.
2. This was intended as a rename of `Task/rejected-*` → `Task/negation-*`. If so, the correct approach is to update the `id` in place on the original entries and keep the original `key` (`Task-rejected-with-code-example` / `Task-rejected-example`), rather than deprecating those keys and re-purposing a different pair. Deprecating the original keys breaks JIRA linkage for any issues logged against them.

Which case applies here?

yanheras:
@lmckenzi This is the initial version (1.0.0-ballot) of this IG. No Jira issues have logged against this IG including these deprecated entries. Those "deprecated" examples were old examples carried over from the QI-Core and our intent is to delete them and define new examples. We would like to remove all of the deprecated entries from this Jira-spec including the two you noted. Since if we were to delete the entries, it will fail the validation, would you please help as the administrator?

These are the four entries that show as deprecated: 
<artifact deprecated="true" id="Task/rejected-with-code-example" key="Task-rejected-with-code-example" name="TaskRejected example rejecting a proposal using a code"/>
<artifact deprecated="true" id="Task/rejected-example" key="Task-rejected-example" name="TaskRejected example rejecting a proposal using a value set"/>
<artifact deprecated="true" id="Encounter/encounter-ed-example" key="Encounter-encounter-ed-example" name="Encounter example of emergency visit that escalated into inpatient patient"/>
<artifact deprecated="true" id="Coverage/example-new" key="Coverage-example-new" name="Coverage example - Subscriber ID"/>

lmckenzi:
**Naming disambiguation.** Even with the old entries kept as deprecated, the surviving active names are essentially identical to the deprecated ones — only a space differs:

     - deprecated `Task/rejected-with-code-example` → *TaskRejected example rejecting a proposal using a code*
     - active `Task/negation-with-code-example` → *Task Rejected example rejecting a proposal using a code*

   (same story for the value-set pair). Consumers browsing the artifact list will see near-duplicates. Could you rename the active `Task/negation-*` entries to something that clearly distinguishes them from the retired pair?

lmckenzi:
@lynnlaakso No need to hold the ballot while waiting for this update.

yanheras:
> **Naming disambiguation.** Even with the old entries kept as deprecated, the surviving active names are essentially identical to the deprecated ones — only a space differs:
> 
> ```
>  - deprecated `Task/rejected-with-code-example` → *TaskRejected example rejecting a proposal using a code*
>  - active `Task/negation-with-code-example` → *Task Rejected example rejecting a proposal using a code*
> ```
> 
> (same story for the value-set pair). Consumers browsing the artifact list will see near-duplicates. Could you rename the active `Task/negation-*` entries to something that clearly distinguishes them from the retired pair?

To make it clear, we do not want to keep those deprecated entries, we request to have the "deprecated" entries physically deleted from the file, there will be no naming disambiguation and near-duplicates. Would you please help delete? The instruction says that to delete entries have to coordinate with the administrator. 

We have a new PR to the IG itself ready to be merged that has those deprecated entries removed https://github.com/HL7/fhir-us-quality-core/pull/40

lmckenzi:
Once stuff is in JiraSpec, it's very hard to remove.  Deprecation is generally the best we can do.  Is renaming possible?

yanheras:
re-open with updated us-quality-core

yanheras:
re-open with updated us-quality-core

PR #1561, opened 2026-08-12
Update FHIR specification and version URLs to "ballot3"

ErikMoll:
see https://chat.fhir.org/#narrow/channel/179192-fmg/topic/Which.20ballot.20version.20should.20assigned.20to.20the.20PoCD.20IG/with/615848130


lmckenzi:
This is failing - you need to fix the PR.  (Ideally run the build process locally before pushing the change.)

PR #1565, opened 2026-08-18
US CQL Common Informative 2

brynrhodes:


lmckenzi (changes requested):
1. `2.0.0-ballot` should be marked `deprecated="true"` now that the official `2.0.0` release is being added. Ballot releases should be deprecated once the corresponding official release ships so people don't submit change requests against them.

2. To confirm on `1.0.0`: with `2.0.0` now the default, would you potentially publish a technical correction against `1.0.0`, or would all further fixes go into the `2.x` line? If the latter, `1.0.0` should also be marked `deprecated="true"`.

brynrhodes:
Agreed, marked both 2.0.0-ballot and 1.0.0 versions as deprecated.

lmckenzi (approved):

PR #1567, opened 2026-08-20
Publishing 2.0.0-snapshot of Vulcan SOA IG

jduteau:


lmckenzi (changes requested):
If you're deprecating 2.0.0, you need to undeprecate 1.0.0

jduteau:
We wouldn't be making any more changes to 1.0 so it is still deprecated.  2.0-snapshot is effectively 2.0 until R6 is officially published.

lmckenzi:
2.0 snapshot isn't an official release though.  The general rule is that 'default' is the most recent official release.  Are you saying that implementers are using it that way?

jduteau:
Yes. Implementers are using our snapshot as an official release.

lmckenzi (approved):

PR #1568, opened 2026-08-23
CQL R2 Spec Artifacts Update

brynrhodes:


lmckenzi (changes requested):
1. **`defaultVersion` should be updated to `2.0.0`.** With the official 2.0.0 release now added, `defaultVersion="1.5.3"` on the `<specification>` element should be bumped to `2.0.0` so the default points at the latest official release rather than a superseded one.

2. **Is 1.5.3 still intended to receive maintenance?** To confirm — do you plan to potentially publish a technical correction against 1.5.3, or is all future work now on the 2.0.x line? If the latter, 1.5.3 should be marked `deprecated="true"` as well so people don't submit change requests against it.

brynrhodes:
I deliberately left the default version as 1.5.3, as that is still the main supported version, it is not deprecated, and if there are issues we absolutely will publish technical corrections against that version. In addition, most users reporting issues will be reporting issues with the current normative publication, as that is by far the more prevalent version in the wild.

PR #1570, opened 2026-08-25
Update FHIR-ebm-incubator.xml

brianalperMD:
update for 1.0.0-snapshot


brianalperMD:
Please note we are wanting to remove the 1.0.0-ballot (noted as the failing check reason) as we did not go to ballot and are publishing a snapshot instead.

lmckenzi (commented):
1. **Citation → PublicationRecord — rename or replace?** The diff is inconsistent. The two example instances (`Citation-306957`, `Citation-179559`) are renamed in place with the original key preserved, which looks right. But everything else is deprecated + re-added under a new key:

   - `StructureDefinition/Citation` → new `StructureDefinition/PublicationRecord` (new key)
   - `List/Citation-examples`, `List/Citation-operations` → new `List/PublicationRecord-*` (new keys)
   - All ~20 `SearchParameter/Citation-*` → new `SearchParameter/PublicationRecord-*` (new keys)
   - `ValueSet/citation-status-type`, `CodeSystem/citation-status-type` → new `publicationrecord-status-type` equivalents (new keys)
   - `Citation/citation-example-research-doi` → new `PublicationRecord/publicationrecord-example-research-doi` (new key)

   If PublicationRecord is conceptually the renamed Citation (same artifact, new name), these should all be updated in place with the original `key` preserved — otherwise JIRA linkage against the existing artifacts is lost. Was the deprecate+add pattern intentional, or should these be renames like the two examples you already handled correctly?

2. **`defaultVersion="current"`.** This was changed from `1.0.0-ballot` to `current` (i.e., the CI build). Should it point to `1.0.0-snapshot` instead, since that's the intended published pre-release version?

3. **`ballotUrl` attribute.** It was updated to `[http://hl7.org/fhir/uv/ebm-incubator/1.0.0-202609-snapshot`](http://hl7.org/fhir/uv/ebm-incubator/1.0.0-202609-snapshot) — but that URL is the snapshot version, not a ballot URL, and you've said you aren't going to ballot. Should the `ballotUrl` attribute be dropped for this release?

PR #1572, opened 2026-08-27
CDC Opioid Data Exchange Profiles 1.0.0

brynrhodes:


lmckenzi (commented):
It looks like this PR's diff against master is only a single whitespace character (removing the space before `/>` on the `1.0.0-ballot` version line). The `1.0.0` publication content — the `1.0.0` version entry, `defaultVersion="1.0.0"`, and `deprecated="true"` on the ballot — is already present on master (added in earlier commits). Did you intend to include additional changes that didn't make it in, or has this PR been superseded and can be closed as redundant?

brynrhodes:
Yes, missed the publication path update, added now

PR #1578, opened 2026-09-10
QM IG UV STU2 and DEQM IG UV STU1

brynrhodes:


lmckenzi (commented):
This is broken.  We can't have the same version number referring to two different releases.  That's not just horrific for implementers, it breaks the tools.  I'm escalating to the TSC.

brynrhodes:
I think the JIRA spec is the thing that's broken, not the specs themselves, there are two different specs, each with different versions:

http://hl7.org/fhir/us/davinci-deqm - Versions 1 - 5 STU
http://hl7.org/fhir/uv/deqm - Version 1

I think what was missed was that the DEQM US versions were included in the ballot, that was incorrect, it should have been a new publication (and there are still two different specifications in JIRA)

lmckenzi:
But it's *NOT* a separate specification.  It's a new version of the same spec whose scope happens to have been broadened.  From an implementer perspective if you've implemented V5 of DaVinci DEQM and you're looking at this new publication it's not "ok, now I need to build a new interface", it's "Here's a few small changes since the last thing I implemented".  The version history would continue to build.  The tracker items surfaced against V3 and V4 continue to be relevant in making decisions for your next release.

